PR #1, opened 2021-04-16
Create a sandboxer binary

vdagonneau:
Create a sandboxer binary similar to the one in the kernel selftests.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,3 +17,8 @@ readme = "README.md"
 [dependencies]
 bitflags = "1.2"
 libc = "0.2"
+
+[dev-dependencies]
+anyhow = "1.0"
+landlock = { path = "." }
+nix = "0.20"
--- a/examples/sandboxer.rs
+++ b/examples/sandboxer.rs
@@ -1,0 +1,151 @@
+use anyhow::bail;
+use landlock::{AccessFs, Ruleset, RulesetAttr};
+use nix::fcntl::{open, OFlag};
+use nix::sys::stat::{fstat, Mode, SFlag};
+use std::env;
+use std::ffi::{OsStr, OsString};
+use std::os::unix::ffi::{OsStrExt, OsStringExt};
+use std::os::unix::process::CommandExt;
+use std::path::PathBuf;
+use std::process::Command;
+
+const ENV_FS_RO_NAME: &str = "LL_FS_RO";
+const ENV_FS_RW_NAME: &str = "LL_FS_RW";
+
+const ACCESS_FS_ROUGHLY_READ: AccessFs = AccessFs::from_bits_truncate(
+    AccessFs::EXECUTE.bits() | AccessFs::READ_FILE.bits() | AccessFs::READ_DIR.bits(),
+);
+
+const ACCESS_FS_ROUGHLY_WRITE: AccessFs = AccessFs::from_bits_truncate(
+    AccessFs::WRITE_FILE.bits()
+        | AccessFs::REMOVE_DIR.bits()
+        | AccessFs::REMOVE_FILE.bits()
+        | AccessFs::MAKE_CHAR.bits()
+        | AccessFs::MAKE_DIR.bits()
+        | AccessFs::MAKE_REG.bits()
+        | AccessFs::MAKE_SOCK.bits()
+        | AccessFs::MAKE_FIFO.bits()
+        | AccessFs::MAKE_BLOCK.bits()
+        | AccessFs::MAKE_SYM.bits(),
+);
+
+const ACCESS_FILE: AccessFs = AccessFs::from_bits_truncate(
+    AccessFs::READ_FILE.bits() | AccessFs::WRITE_FILE.bits() | AccessFs::EXECUTE.bits(),
+);
+
+/// Populates a given ruleset with PathBeneath landlock rules
+///
+/// # Arguments
+///
+/// * `ruleset` - The ruleset to add the rules to. Note that due to the current
+///   API, it is also returned at the end.
+/// * `paths` - An OsString that contains the paths that are going to be
+///   restricted. Paths are separated with ":", e.g. "/bin:/lib:/usr:/proc". In
+///   case an empty string is provided, NO restrictions are applied.
+/// * `access` - The set of restrictions to apply to each of the given paths.
+///
+fn populate_ruleset(
+    mut ruleset: Ruleset,
+    paths: OsString,
+    access: AccessFs,
+) -> Result<Ruleset, anyhow::Error> {
+    if paths.len() == 0 {
+        return Ok(ruleset);
+    }
+
+    let individual_paths: Vec<PathBuf> = paths
+        .into_vec()
+        .split(|b| *b == b':')
+        .map(|v| OsStr::from_bytes(v).to_owned().into())
+        .collect();
+
+    for path in individual_paths {
+        match open(&path, OFlag::O_PATH | OFlag::O_CLOEXEC, Mode::empty()) {
+            Err(e) => {
+                bail!("Failed to open \"{}\": {}", path.to_string_lossy(), e);
+            }
+            Ok(parent) => match fstat(parent) {
+                Ok(stat) => {
+                    let actual_access =
+                        if (stat.st_mode & SFlag::S_IFMT.bits()) != SFlag::S_IFDIR.bits() {
+                            access & ACCESS_FILE
+                        } else {
+                            access
+                        };
+
+                    ruleset = match ruleset.add_path_beneath_rule(parent, actual_access) {
+                        Ok(r) => r,
+                        Err(e) => {
+                            bail!(
+                                "Failed to update ruleset with \"{}\": {}",
+                                path.to_string_lossy(),
+                                e
+                            );
+                        }
+                    };
+                }
+                Err(e) => {
+                    bail!("Failed to stat \"{}\": {}", path.to_string_lossy(), e);
+                }
+            },
+        }
+    }
+
+    Ok(ruleset)
+}
+
+fn main() -> Result<(), anyhow::Error> {
+    let args: Vec<_> = env::args_os().collect();
+
+    let program_name = args.get(0).map(|s| s.to_string_lossy()).unwrap_or_default();
+
+    if args.len() < 2 {
+        eprintln!(
+            "usage: {}=\"...\" {}=\"...\" {} <cmd> [args]...\n",
+            ENV_FS_RO_NAME, ENV_FS_RW_NAME, program_name
+        );
+        eprintln!("Launch a command in a restricted environment.\n");
+        eprintln!("Environment variables containing paths, each separated by a colon:");
+        eprintln!(
+            "* {}: list of paths allowed to be used in a read-only way.",
+            ENV_FS_RO_NAME
+        );
+        eprintln!(
+            "* {}: list of paths allowed to be used in a read-write way.",
+            ENV_FS_RW_NAME
+        );
+        eprintln!(
+            "\nexample:\n\
+                {}=\"/bin:/lib:/usr:/proc:/etc:/dev/urandom\" \
+                {}=\"/dev/null:/dev/full:/dev/zero:/dev/pts:/tmp\" \
+                {} bash -i",
+            ENV_FS_RO_NAME, ENV_FS_RW_NAME, program_name
+        );
+        return Ok(());
+    }
+
+    let fs_ro = env::var_os(ENV_FS_RO_NAME)
+        .expect(&format!("Missing environment variable {}", ENV_FS_RO_NAME));
+    let fs_rw = env::var_os(ENV_FS_RW_NAME)
+        .expect(&format!("Missing environment variable {}", ENV_FS_RW_NAME));
+
+    let cmd_name = args.get(1).map(|s| s.to_string_lossy()).unwrap();
+
+    let ruleset = RulesetAttr::new().handle_fs(AccessFs::all()).create()?;
+    let ruleset = populate_ruleset(ruleset, fs_ro, ACCESS_FS_ROUGHLY_READ)?;
+    populate_ruleset(
+        ruleset,
+        fs_rw,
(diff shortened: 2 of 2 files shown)


crates/sandboxer/main.rs
```diff
@@ -0,0 +1,152 @@
...
+    if let Err(_) = ruleset.restrict_self() {
+        eprintln!("Failed to enforce ruleset");
+        return;
+    }
+
+    Command::new(cmd_name.to_string())
+        .env_remove(ENV_FS_RO_NAME)
+        .env_remove(ENV_FS_RW_NAME)
+        .args(env::args().skip(2))
+        .exec();
```
l0kod, line 151: `exec()` returns an `Error`, it should be handled.

Cargo.toml
```diff
@@ -1,19 +1,29 @@
...
 homepage = "https://landlock.io"
 repository = "https://github.com/landlock-lsm/rust-landlock"
 license = "MIT OR Apache-2.0"
 keywords = ["access-control", "linux", "sandbox", "security"]
 exclude = [".gitignore"]
 readme = "README.md"
 
+[[bin]]
+bench = false
+path = "crates/sandboxer/main.rs"
```
l0kod, line 19: Let's simply put it in `examples/sandboxer.rs` for now and not move all other files. It could then be run with `cargo run --example=sandboxer`.
We can also add the required dependencies with `[dev-dependencies]`.

crates/sandboxer/main.rs
```diff
@@ -0,0 +1,152 @@
...
+            "* {}: list of paths allowed to be used in a read-only way.",
+            ENV_FS_RO_NAME
+        );
+        eprintln!(
+            "* {}: list of paths allowed to be used in a read-write way.",
+            ENV_FS_RW_NAME
+        );
+        eprintln!(
+            "\nexample:\n\
+				     {}=\"/bin:/lib:/usr:/proc:/etc:/dev/urandom\" \
```
l0kod, line 104: There is tabs mixed with spaces. Let's use spaces only!

crates/sandboxer/main.rs
```diff
@@ -0,0 +1,152 @@
...
+        .expect(&format!("Missing environment variable {}", ENV_FS_RO_NAME));
+    let fs_rw = env::var_os(ENV_FS_RW_NAME)
+        .expect(&format!("Missing environment variable {}", ENV_FS_RW_NAME));
+
+    let cmd_name = args
+        .get(1)
+        .map(|s| s.to_string_lossy())
+        .unwrap_or(Cow::Borrowed("bash -i"));
+
+    let ruleset = match RulesetAttr::new().create() {
```
l0kod, line 122: This will failed because there is no handled accesses. You need to call `handle_fs()`.

crates/sandboxer/main.rs
```diff
@@ -0,0 +1,152 @@
...
+        Err(_) => {
+            eprintln!("Failed to create a ruleset");
+            return;
+        }
+        Ok(ruleset) => ruleset,
+    };
+
+    let ruleset = match populate_ruleset(ruleset, fs_ro, ACCESS_FS_ROUGHLY_READ) {
```
l0kod, line 130: It would be cleaner to use `?` and then convert `main()` to return a `Result`: https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/question-mark-in-main-and-tests.html

crates/sandboxer/main.rs
```diff
@@ -0,0 +1,152 @@
...
+        Ok(r) => r,
+        Err(_) => return,
+    };
+
+    let ruleset = match populate_ruleset(ruleset, fs_rw, ACCESS_FS_ROUGHLY_READ | ACCESS_FS_ROUGHLY_WRITE) {
```
l0kod, line 135: It would be nice to avoid repeated `let ruleset` but it is not possible with the current API.

l0kod (commented):
Nice! Here are some comments.

Please use a commit message starting with a "sandboxer: " prefix and a Signed-off-by tag.

You would then need to run `cargo fmt` and test it.

Cargo.lock
```diff
@@ -0,0 +1,55 @@
+# This file is automatically @generated by Cargo.
```
l0kod, line 1: We should not commit `Cargo.lock` for now.

Cargo.toml
```diff
@@ -6,7 +6,7 @@ authors = [
 	"Vincent Dagonneau <vincentdagonneau@gmail.com>"
 ]
 edition = "2018"
-description = "Landlock LSM helpers"
+description = "Landlock LSM sandboxer"
```
l0kod, line 9: Don't need to change this.

Cargo.toml
```diff
@@ -17,3 +17,8 @@ readme = "README.md"
 [dependencies]
 bitflags = "1.2"
 libc = "0.2"
+
+[dev-dependencies]
+anyhow = "1.0.40"
+nix = "0.20.0"
+landlock = { version = "0.1.0", path = "." }
```
l0kod, line 24: No need to pin the Landlock version.

l0kod (commented):
The Linux version does handle empty environment variables, which is handy. Could you please handle this use case as well?

examples/sandboxer.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    let fs_ro = env::var_os(ENV_FS_RO_NAME)
+        .expect(&format!("Missing environment variable {}", ENV_FS_RO_NAME));
+    let fs_rw = env::var_os(ENV_FS_RW_NAME)
+        .expect(&format!("Missing environment variable {}", ENV_FS_RW_NAME));
+
+    let cmd_name = args
+        .get(1)
+        .map(|s| s.to_string_lossy())
+        .unwrap_or(Cow::Borrowed("bash -i"));
```
l0kod, line 122: This is unfortunately not reached because the `args.len()` check error out before.

vdagonneau:
> The Linux version does handle empty environment variables, which is handy. Could you please handle this use case as well?

I am not sure what you mean. If I understand correctly, you are referring to this part in the original Linux version: [sandboxer.c:98](https://github.com/landlock-lsm/linux/commit/e3f0859b73352f2beae6eae1ecd356355137f644#diff-d57a3a6675aea6e1eefd15ad854d8f416fa334eae25cd7b379fc5cf3d5f9506eR98). My understanding is that in case of an empty environment variable, you return early from `populate_ruleset`, not applying any restrictions.

In the Rust version, the `for path in individual_paths` loop takes care of an empty environment variable as it does not enter the loop at all and returns right after.

examples/sandboxer.rs
vdagonneau: I am not sure I understand, `cmd_name` here is the command requested by the user e.g. `bash`. It corresponds to [sandboxer.c:226](https://github.com/landlock-lsm/linux/commit/e3f0859b73352f2beae6eae1ecd356355137f644#diff-d57a3a6675aea6e1eefd15ad854d8f416fa334eae25cd7b379fc5cf3d5f9506eR226).

examples/sandboxer.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    Ok(ruleset)
+}
+
+fn main() -> Result<(), anyhow::Error> {
+    let args: Vec<_> = env::args_os().collect();
+
+    let program_name = args.get(0).map(|s| s.to_string_lossy()).unwrap_or_default();
+
+    if args.len() < 2 {
```
l0kod, line 89: This check makes `unwrap_or(Cow::Borrowed("bash -i"))` useless, we still require to set a command.

l0kod:
> I am not sure what you mean. If I understand correctly, you are referring to this part in the original Linux version: [sandboxer.c:98](https://github.com/landlock-lsm/linux/commit/e3f0859b73352f2beae6eae1ecd356355137f644#diff-d57a3a6675aea6e1eefd15ad854d8f416fa334eae25cd7b379fc5cf3d5f9506eR98). My understanding is that in case of an empty environment variable, you return early from `populate_ruleset`, not applying any restrictions.

Not exacly, it is required to set the environement variable but it can be empty. The `num_paths == 1` is misleading but it means that there is one entry and the `path_list[0][0] == '\0'` checks that this entry is empty.

examples/sandboxer.rs
l0kod: `cmd_name` is correct, I was talking about the default value (cf. other comment).

examples/sandboxer.rs
```diff
@@ -0,0 +1,154 @@
...
+    AccessFs::READ_FILE.bits() | AccessFs::WRITE_FILE.bits() | AccessFs::EXECUTE.bits(),
+);
+
+/// Populates a given ruleset with PathBeneath landlock rules
+///
+/// # Arguments
+///
+/// * `ruleset` - The ruleset to add the rules to. Note that due to the current
+///   API, it is also returned at the end.
+/// * `paths` - An OsString that contains the paths taht are going to be
```
l0kod, line 45: Typo: "taht"

examples/sandboxer.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    for path in individual_paths {
+        match open(&path, OFlag::O_PATH | OFlag::O_CLOEXEC, Mode::empty()) {
+            Err(e) => {
+                bail!("Failed to open \"{}\": {}\n", path.to_string_lossy(), e);
+            }
+            Ok(parent_fd) => match fstat(parent_fd) {
+                Ok(stat) => {
+                    if (stat.st_mode & SFlag::S_IFMT.bits()) != SFlag::S_IFDIR.bits() {
+                        actual_access.insert(ACCESS_FILE);
```
l0kod, line 76: There is an issue with the logic here. It doesn't work with files:
```
LL_FS_RO="/bin:/lib:/usr:/proc:/etc:/dev/urandom" LL_FS_RW="/dev/null:/dev/full:/dev/zero:/dev/pts:/tmp" target/debug/examples/sandboxer bash -i
file: "/dev/urandom"
Error: Failed to update ruleset with "/dev/urandom": Invalid argument (os error 22)

```
An extra carriage return is also appended at the end of an error message.

examples/sandboxer.rs
```diff
@@ -0,0 +1,154 @@
...
+);
+
+/// Populates a given ruleset with PathBeneath landlock rules
+///
+/// # Arguments
+///
+/// * `ruleset` - The ruleset to add the rules to. Note that due to the current
+///   API, it is also returned at the end.
+/// * `paths` - An OsString that contains the paths taht are going to be
+///   restricted. Paths are separated with ":" e.g. "/bin:/lib:/usr:/proc". In
```
l0kod, line 46: You can add a comma before "e.g.".

l0kod (changes requested):
Great! This should be the last fixes.

examples/sandboxer.rs
```diff
@@ -0,0 +1,154 @@
+use anyhow::bail;
+use landlock::{AccessFs, Ruleset, RulesetAttr};
+use nix::fcntl::{open, OFlag};
+use nix::sys::stat::{fstat, Mode, SFlag};
+use std::os::unix::ffi::{OsStrExt, OsStringExt};
+use std::os::unix::process::CommandExt;
+use std::process::Command;
+use std::{
+    borrow::Cow,
```
l0kod, line 9: `Cow` is not used anymore.

examples/sandboxer.rs
```diff
@@ -0,0 +1,153 @@
...
+///
+/// * `ruleset` - The ruleset to add the rules to. Note that due to the current
+///   API, it is also returned at the end.
+/// * `paths` - An OsString that contains the paths that are going to be
+///   restricted. Paths are separated with ":", e.g. "/bin:/lib:/usr:/proc". In
+///   case an empty string is provided, NO restrictions are applied.
+/// * `access` - The set of restrictions to apply to each of the given paths.
+///
+fn populate_ruleset(
+    ruleset: Ruleset,
```
l0kod, line 50: It makes sense to write `mut ruleset`…

examples/sandboxer.rs
```diff
@@ -0,0 +1,153 @@
...
+    paths: OsString,
+    access: AccessFs,
+) -> Result<Ruleset, anyhow::Error> {
+    let mut ruleset = ruleset;
```
l0kod, line 54: …instead of this line.

examples/sandboxer.rs
```diff
@@ -0,0 +1,153 @@
...
+    let mut actual_access = access.clone();
```
l0kod, line 55: Same here, you can simply replace `actual_access` with a `mut access`.

examples/sandboxer.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    for path in individual_paths {
+        match open(&path, OFlag::O_PATH | OFlag::O_CLOEXEC, Mode::empty()) {
+            Err(e) => {
+                bail!("Failed to open \"{}\": {}", path.to_string_lossy(), e);
+            }
+            Ok(parent_fd) => match fstat(parent_fd) {
+                Ok(stat) => {
+                    if (stat.st_mode & SFlag::S_IFMT.bits()) != SFlag::S_IFDIR.bits() {
+                        actual_access &= ACCESS_FILE;
```
l0kod, line 75: There is a bug here because access rights are modified for all following paths in the loop.

examples/sandboxer.rs
l0kod: In fact, no, we need to add a new variable for each loop walk.

examples/sandboxer.rs
vdagonneau: Nice catch, pretty sneaky bug.

l0kod: `mut` can lead to bugs. :wink:

l0kod:
I just added some cosmetic changes:
- shorten crate versions (dev-dependencies)
- flatten `use` lines
- removed useless `mut ruleset`
- removed useless `let ruleset`
- ran `cargo fmt`
- improved commit description

l0kod:
Thanks!

PR #3, opened 2021-04-24
Create a compatible-friendly API

l0kod:
This will enable easy error handling for backward and forward compatibility, both from the kernel ABI and the crate user point of view. The goal is to encourage best-effort security but still enable to handle partial incompatibility.

### DIFF
--- a/examples/sandboxer.rs
+++ b/examples/sandboxer.rs
@@ -1,5 +1,5 @@
 use anyhow::{anyhow, bail};
-use landlock::{AccessFs, Compat, ErrorThreshold, PathBeneath, Ruleset, RulesetAttr};
+use landlock::{AccessFs, Compat, ErrorThreshold, Factory, Ruleset};
 use nix::fcntl::{open, OFlag};
 use nix::sys::stat::{fstat, Mode, SFlag};
 use std::env;
@@ -45,6 +45,7 @@ const ACCESS_FILE: AccessFs = AccessFs::from_bits_truncate(
 /// * `access` - The set of restrictions to apply to each of the given paths.
 ///
 fn populate_ruleset(
+    factory: &Factory,
     ruleset: Compat<Ruleset>,
     paths: OsString,
     access: AccessFs,
@@ -74,7 +75,9 @@ fn populate_ruleset(
                         Ok(inner_ruleset
                             .set_error_threshold(ErrorThreshold::PartiallyCompatible)
                             .add_rule(
-                                PathBeneath::new()?
+                                // PathBeneath
+                                factory
+                                    .create()?
                                     .set_parent(&parent)?
                                     .allow_access(actual_access)?,
                             )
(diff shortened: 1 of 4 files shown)


vdagonneau:
I did not have much time to review over the weekend, it is going to take me more time to wrap my head around this code.

I had one question though: wouldn't it make sense at this point to implement a proper error type for the crate that maps to the exact errors we get from the landlock syscalls. I feel like using std::io::error does not map cleanly to the semantics of the sycalls.

l0kod:
> I had one question though: wouldn't it make sense at this point to implement a proper error type for the crate that maps to the exact errors we get from the landlock syscalls. I feel like using std::io::error does not map cleanly to the semantics of the sycalls.

The Landlock syscalls can return multiple errors (e.g. ENOMEM, EBADF…) as every syscall, and because of kernel call chains, these errors could evolve with new kernel releases. The `Compat<T>` wrapper should make EINVAL errors silent (if not explicitly managed with `.into_result()`) to build a best-effort sandboxing. Some other errors like EBADFD should not happen because of Rust traits (e.g. `AsRawFd`), but in practice it is still possible (e.g. if a safe types wrapping unsafe ones have bugs). It would make sense to use [Nix's Errno](https://docs.rs/nix/0.20.0/nix/errno/enum.Errno.html) but I think we should avoid pulling dependencies when possible. On the other side, the Landlock (raw) syscalls will be implemented in Nix eventually…

Overall, I think it would make sense to have errno wrapping for Landlock syscalls. It would help to have a custom error type able to store some Rule properties like file name. I was also planning to add something like `add_rule_iter(Iterator<(Rule, ErrorThreshold)>)`, which would make this kind of custom error type more interesting.

So yes, let's think about one or multiple Landlock error types (without Nix dependency for now).

l0kod:
Changes:
* replace `let Some()` with `match`
* cleanup `Compat<Ruleset>::create()`
* improve doc
* fix `self.set_last_call_status(LastCall::Fake)`

l0kod:
Changes:
* fix restrict_self() return for None
* use map_or()

l0kod:
Changes:
* fix `LastCall::Fake` handling
* add new commit to better handle build chain errors

l0kod:
Changes:
* Simplify `.into_result()`
* Revamp the last commit to create a proper `Factory`.

vdagonneau:
If I understand the code well, the current factory is essentially there to hold the abi_version value, it caches that value and allows the user to create multiple objects without having to call LANDLOCK_CREATE_RULESET_VERSION each time.

Since the landlock abi version is essentially static for the entirety of the programs execution (barring exotic setups) wouldn't it make sense and be simpler to just use a global mutable abi_version variable? This variable would be set the first time the landlock interface is called and would subsequently be available for the other calls.

I know "global mutable state" sounds ... not great. However, couldn't we get away in this case with using something like an atomic: https://doc.rust-lang.org/std/sync/atomic/ ?

l0kod:
> If I understand the code well, the current factory is essentially there to hold the abi_version value, it caches that value and allows the user to create multiple objects without having to call LANDLOCK_CREATE_RULESET_VERSION each time.

Right.

> Since the landlock abi version is essentially static for the entirety of the programs execution (barring exotic setups) wouldn't it make sense and be simpler to just use a global mutable abi_version variable? This variable would be set the first time the landlock interface is called and would subsequently be available for the other calls.
> 
> I know "global mutable state" sounds ... not great. However, couldn't we get away in this case with using something like an atomic: https://doc.rust-lang.org/std/sync/atomic/ ?

Great idea! I was thinking about a `RefCell` but `Atomicl32` is enough for this kind of cache (which can be implemented without lock).

l0kod:
> If I understand the code well, the current factory is essentially there to hold the abi_version value, it caches that value and allows the user to create multiple objects without having to call LANDLOCK_CREATE_RULESET_VERSION each time.

Thinking more about that, I'm now a bit reluctant. For now `Factory` doesn't have a `set_error_threshold()` method, but this would be handy for users that want to make sure that all Landlock features are supported by the running system. Removing the `Factory` would make this kind of error handling more verbose and error-prone because the crate user would need to not forget to set this property for all new objects (e.g. `PathBeneath`).

An alternative approach would be to pass this kind of `Factory/Setting` as argument to each object creation. It is more verbose but could be less confusing.

Another alternative would be to handle this error threshold globally, but I don't think it is a wise idea.

l0kod:
Changes:
* Implement the first variant: replace `Factory` with `Compatibility` object

l0kod:
Let's use this version for now. We'll update it incrementally if needed.

PR #4, opened 2021-04-29
profile: Added comments and a finally functioning profile loader.

vdagonneau:
This first version is really basic. It just loads a profile based on the
command name located in ~/.config/file_<cmd>.ll.

The file format is also very basic, it is essentially just like dotenv
files: a succession of environment variable declaration. The only thing
"smart" about it for now is that it merges them before passing them to
the sandboxer.

Note: I included code in the comments much like for a normal crate.
However, it seems like code in the comments of examples are not
currently being executed anywhere: https://github.com/rust-lang/cargo/issues/4508

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,3 +22,4 @@ libc = "0.2"
 anyhow = "1.0"
 landlock = { path = "." }
 nix = "0.20"
+nom = "6.1"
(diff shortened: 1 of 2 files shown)


l0kod (commented):
I think profile definitions deserve a proper configuration file format. We also need parsing and functional tests, as unit tests for functions but also for full profiles. This applies to the crate and this example.

l0kod:
This new application should probably be hosted in a dedicated repository. I'll probably keep the sandboxer example as an example that can be used to help users, but let's focus on the original goal: a library. :)

PR #5, opened 2021-08-03
Replace bitflags with custom Access* types

l0kod:
This is the continuation of #3 

The bitflags implementation makes it possible for developers to set
arbitrary and potentially unknown access rights without being noticed.

Replace bitflags with a custom AccessFs and AccessRighs<AccessFs> types,
and AccessFlags, AccessFlagsInner and AccessFlagsFs traits to get a
deterministic API and avoid type confusion issues.  Add a set of
pre-defined filesystem access rights (read, make, remove and group1) to
make it easier to use these rights.  This new types will also enable to
add runtime compatibility checks.

Because it is not possible to implement const trait yet, we can't set
const access right.  See
https://github.com/rust-lang/rfcs/blob/master/text/0911-const-fn.md

Add fake network access rights for test purpose, to check that this
approach is generic enough and guard against type confusion issues for
future access right types.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,6 @@ exclude = [".gitignore"]
 readme = "README.md"
 
 [dependencies]
-bitflags = "1.2"
 libc = "0.2"
 
 [dev-dependencies]
--- a/examples/sandboxer.rs
+++ b/examples/sandboxer.rs
@@ -1,6 +1,7 @@
 use anyhow::{anyhow, bail};
 use landlock::{
-    AccessFs, Compat, Compatibility, ErrorThreshold, PathBeneath, RulesetCreated, RulesetInit,
+    AccessFs, AccessRights, Compat, Compatibility, ErrorThreshold, PathBeneath, RulesetCreated,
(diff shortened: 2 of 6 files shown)


gnoack:
related to https://github.com/landlock-lsm/go-landlock/issues/15, where I contemplated the same for go-landlock. (I ended up not doing it there, as wrapping that uint64 type had complicated the API quite a bit. So in the end, in Go, we're just checking validity of these arguments when they are passed to `NewConfig()` and `RestrictPaths()`.

l0kod:
I'm still unsure about the best approach… I agree that it can add more complexity. I may not replace this type after all but maybe only implement a trait for it.

I wanted at least to check the value of these bits when they are used, not against a static bitmask but against the actual accesses supported by the running kernel (see [Compatibility](https://github.com/landlock-lsm/rust-landlock/blob/060ce63987ae40fe387fe642e70a83240ae20992/src/lib.rs#L223)). But thanks to the builder pattern, we can easily check if a set of accesses is OK or not at runtime. :thinking:

PR #6, opened 2021-10-12
Revamp the compatibility API

l0kod:
This is the continuation of #3 and #5.

This is a lot of changes to come from the previous complex and incomplete implementation (#3) to a new one, simpler, cleaner, easier to use and complete. It should be much easier to review each commit instead of the whole PR changes.

There is two main goals:
* Make the compatibility mechanism simpler for this library users and developers.
* Only avoid useless Landlock specific calls but run as much code as possible in any case: supported system or not.

# Rational for the current interface

The simple approach is to expose the Landlock ABI version and ask the developer to manage the related implication explicitly. This requires developers to grasp all the potential behavior changes over time, which will eventually lead to issues.

Instead, we can ask developers to identify optional (best-effort) or required features. This make more sense because this directly relies on the access-control semantic. It makes the library more difficult to implement but simpler to use for most use cases (because there is no manual ABI version management), and easier to audit for most complex requirements (because they are explicitly set as required, instead on relying on manual ABI version mapping).

See more details in the updated [README.md](https://github.com/landlock-lsm/rust-landlock/blob/features/compat2/README.md#compatibility).


l0kod:
I have taken into account comment https://github.com/landlock-lsm/rust-landlock/commit/d542cd1c5f314b15c5f929aadbfb8d48be3f61c3#r58020924 (not visible anymore because of the rebase) with 0ea634c27cc8378cfbadd8716902671cb585b85a.

l0kod:
Thanks @gnoack!

Feel free to open an issue for further comments @vdagonneau.

PR #8, opened 2021-10-20
Add high-level errors

l0kod:
Replace io::Error with high-level custom errors and fix some issues. This depends on #6 and #7.

### DIFF
--- a/src/compat.rs
+++ b/src/compat.rs
@@ -57,8 +57,6 @@ fn abi_from() {
 #[cfg_attr(test, derive(Debug, PartialEq))]
 #[derive(Copy, Clone)]
 pub(crate) enum CompatState {
-    /// Initial unknown state.
-    Start,
     /// All requested restrictions are enforced.
     Full,
     /// Some requested restrictions are enforced, following a best-effort approach.
@@ -74,8 +72,6 @@ impl CompatState {
         *self = match (*self, other) {
             (CompatState::Final, _) => CompatState::Final,
             (_, CompatState::Final) => CompatState::Final,
-            (CompatState::Start, state) => state,
-            (state, CompatState::Start) => state,
             (CompatState::No, CompatState::No) => CompatState::No,
             (CompatState::Full, CompatState::Full) => CompatState::Full,
             (_, _) => CompatState::Partial,
@@ -85,21 +81,18 @@ impl CompatState {
 
 #[test]
 fn compat_state_update_1() {
-    let mut state = CompatState::Start;
+    let mut state = CompatState::Full;
 
-    state.update(CompatState::Start);
-    assert_eq!(state, CompatState::Start);
+    state.update(CompatState::Full);
+    assert_eq!(state, CompatState::Full);
 
     state.update(CompatState::No);
-    assert_eq!(state, CompatState::No);
-
-    state.update(CompatState::Start);
-    assert_eq!(state, CompatState::No);
+    assert_eq!(state, CompatState::Partial);
 
     state.update(CompatState::Full);
     assert_eq!(state, CompatState::Partial);
 
-    state.update(CompatState::Start);
+    state.update(CompatState::Full);
     assert_eq!(state, CompatState::Partial);
 
     state.update(CompatState::No);
@@ -109,9 +102,6 @@ fn compat_state_update_1() {
     assert_eq!(state, CompatState::Final);
 
     state.update(CompatState::Full);
-    assert_eq!(state, CompatState::Final);
-
-    state.update(CompatState::Start);
     assert_eq!(state, CompatState::Final);
 }
 
@@ -125,7 +115,7 @@ fn compat_state_update_2() {
     state.update(CompatState::No);
     assert_eq!(state, CompatState::Partial);
 
-    state.update(CompatState::Start);
+    state.update(CompatState::Full);
     assert_eq!(state, CompatState::Partial);
 }
 
@@ -156,7 +146,7 @@ impl From<ABI> for Compatibility {
             state: match abi {
                 // Forces the state as unsupported because all possible types will be useless.
                 ABI::Unsupported => CompatState::Final,
-                _ => CompatState::Start,
+                _ => CompatState::Full,
             },
         }
     }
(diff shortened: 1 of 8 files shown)


l0kod:
Any though @vdagonneau?

l0kod:
I'm a bit hesitant about the `HandleAccesses`, `HandleAccessesError` and `HandleAccessError`-like naming but I didn't find better ones for now.

src/compat.rs
```diff
@@ -172,102 +165,109 @@ pub trait Compatible {
...
+    T: Access,
 {
-    fn try_compat(self, compat: &mut Compatibility) -> Result<Self, Error> {
-        let (state, ret) = if self.is_empty() {
+    unsafe { BitFlags::<T>::from_bits_unchecked(!flags.bits()) }
+}
+
+#[test]
+fn bit_flags_full_negation() {
+    let scopped_negation = !BitFlags::<AccessFs>::all();
```
vdagonneau, line 185: nitpick: I think scoped is the right spelling here.

vdagonneau:
I realize that what I'm going to suggest is *NOT* part of this PR but ...

I am not sure the use of `set_best_effort` is super clear for the end-user. The Ruleset object is essentially a builder pattern that eventually leads to a create that actually builds the `RulesetCreated` object. That is clear. However, I'm not sure I have ever seen a builder with `set` methods used to manipulate transient variables that are not relevant to the final object. Here the `set_best_effort` role is to alter the behavior of further calls to `handle_fs`.

I think a different way to do it would be to either pass `is_best_effort` to `handle_fs` or to have two 'flavors' of `handle_fs`: `handle_fs_strict` and `handle_fs_best_effort`. Either way since best effort is an integral part of the context of the `handle_fs` method, it should be expressed that way.

The other thing that I think is not super clear although pretty smart is the Compatible trait (and therefore the calls to set_best_effort). It was definitely not obvious to me that I could use set_best_effort both on the entire Ruleset and on the individual Rule (PathBeneath) later on. I do understand the reason but I think that might be ... too ... clever ;-). Besides the remark above kinda applies to set_best_effort on rules as well. 

For reference, this is the bit of code that took me a long time to figure out:

```rust
Ruleset::new()
            // Must have at least the execute check…
            .set_best_effort(false)
            .handle_fs(AccessFs::Execute)?
            // …and possibly others (superset of AccessFs::Execute).
            .set_best_effort(true)
            .handle_fs(ABI::V1)?
            .create()?
            .set_no_new_privs(true)
            .add_rule(PathBeneath::new(&PathFd::new("/")?).allow_access(ABI::V1))?
            .restrict_self()
```

Something simpler:
```rust
Ruleset::new()
            // Must have at least the execute check…
            .handle_fs_strict(AccessFs::Execute)?
            // …and possibly others (superset of AccessFs::Execute).
            .handle_fs(ABI::V1)?
            .create()?
            .set_no_new_privs(true)
            .add_rule(PathBeneath::new(&PathFd::new("/")?).allow_access(ABI::V1))?
            .restrict_self()
```

I think there is also more simplification to be done, for example, I don't like the fact that `handle_fs` returns a Result. Unless I missed something it should be possible to 'constify' try_compat so that we do not shift the burden of checking to the programmer at runtime. Take that with a grain of salt though because I think that is not an easy task and I'm not even sure if that is possible today. But that would be pretty nice though ;-)

l0kod:
> I am not sure the use of `set_best_effort` is super clear for the end-user. The Ruleset object is essentially a builder pattern that eventually leads to a create that actually builds the `RulesetCreated` object. That is clear. However, I'm not sure I have ever seen a builder with `set` methods used to manipulate transient variables that are not relevant to the final object. Here the `set_best_effort` role is to alter the behavior of further calls to `handle_fs`.
> 
> I think a different way to do it would be to either pass `is_best_effort` to `handle_fs` or to have two 'flavors' of `handle_fs`: `handle_fs_strict` and `handle_fs_best_effort`. Either way since best effort is an integral part of the context of the `handle_fs` method, it should be expressed that way.

The issue is that the best-effort approach is applied to other things that `Ruleset::handle_access()` (renamed from `handle_fs`): `RulesetCreated::add_rule()` and `PathBeneath`. I would like to keep the interface as simple as possible and not verbose for a common use (i.e. best-effort approach).

> The other thing that I think is not super clear although pretty smart is the Compatible trait (and therefore the calls to set_best_effort). It was definitely not obvious to me that I could use set_best_effort both on the entire Ruleset and on the individual Rule (PathBeneath) later on. I do understand the reason but I think that might be ... too ... clever ;-). Besides the remark above kinda applies to set_best_effort on rules as well.

It might be a bit hidden behind the Compatible trait, but `set_best_effort()` should only be called by developers that know what they are doing anyway. ;)

> For reference, this is the bit of code that took me a long time to figure out:
> 
> ```rust
> Ruleset::new()
>             // Must have at least the execute check…
>             .set_best_effort(false)
>             .handle_fs(AccessFs::Execute)?
>             // …and possibly others (superset of AccessFs::Execute).
>             .set_best_effort(true)
>             .handle_fs(ABI::V1)?
>             .create()?
>             .set_no_new_privs(true)
>             .add_rule(PathBeneath::new(&PathFd::new("/")?).allow_access(ABI::V1))?
>             .restrict_self()
> ```
> 
> Something simpler:
> 
> ```rust
> Ruleset::new()
>             // Must have at least the execute check…
>             .handle_fs_strict(AccessFs::Execute)?
>             // …and possibly others (superset of AccessFs::Execute).
>             .handle_fs(ABI::V1)?
>             .create()?
>             .set_no_new_privs(true)
>             .add_rule(PathBeneath::new(&PathFd::new("/")?).allow_access(ABI::V1))?
>             .restrict_self()
> ```

I have mitigated feelings about this approach. It is indeed simpler to understand/review but less generic. It should be noted that this example is an (extreme) test case, but it should not be common to use this construct.

> I think there is also more simplification to be done, for example, I don't like the fact that `handle_fs` returns a Result. Unless I missed something it should be possible to 'constify' try_compat so that we do not shift the burden of checking to the programmer at runtime. Take that with a grain of salt though because I think that is not an easy task and I'm not even sure if that is possible today. But that would be pretty nice though ;-)

I would like that, but unfortunately this is not possible, e.g. when passing an empty `BitFlags<AccessFs>` set, or when it contains unknown accesses, which may not be known at compile time.

l0kod:
I merged these important changes but we can continue the discussion here or in a following commit adding a proper documentation.

l0kod:
See #9.

PR #9, opened 2021-11-25
Add proper documentation

l0kod:


l0kod:
@vdagonneau and @gnoack, could you please take a look?

src/lib.rs
```diff
@@ -1,3 +1,75 @@
...
+//! This kind of sandbox is expected to help mitigate the security impact of bugs,
+//! unexpected or malicious behaviors in applications.
+//! Landlock empowers any process, including unprivileged ones, to securely restrict themselves.
+//! More information about Landlock can be found in the [official website](https://landlock.io).
+//!
+//! This crate provides a safe abstraction for the Landlock system calls, along with some helpers.
+//!
+//! # Use cases
+//!
+//! This crate is especially useful for build-in application sandboxing:
```
gnoack, line 15: typo: s/build-in/built-in/

src/lib.rs
```diff
@@ -1,3 +1,75 @@
...
+//! * Parsers hardening (e.g., archive tools, file format conversion, renderers).
```
gnoack, line 16: *nit* I would probably say "parser hardening" here (without the "s")

src/lib.rs
```diff
@@ -1,3 +1,75 @@
...
+//! * (Part of) applications with limited file renaming or linking needs
+//!   (e.g., some system or network services).
+//! * Monolith applications dealing with different levels of confidentiality
+//!   (e.g., web browser, email server).
```
gnoack, line 20: +1, good examples; (web browsers already use seccomp, and I believe qmail has such a compartmentalized architecture); Some terminlogogy nitpicking: I'm not sure if these still qualify as monoliths though... it's a loaded term, and actually to use process-level sandboxing, the design often needs to be split into multiple processes with different permissions (like qmail)

src/lib.rs
```diff
@@ -1,3 +1,75 @@
...
+//!
+//! # Examples
+//!
+//! Examples can be found with the [`Ruleset` documentation](Ruleset)
+//! and the [sandboxer example](https://github.com/landlock-lsm/rust-landlock/blob/master/examples/sandboxer.rs).
+//!
+//! # Current limitations
+//!
+//! This crate handles the Landlock features available since Linux 5.13
```
gnoack, line 29: Suggestion:

> This crate *exposes* the Landlock features available *as of* Linux 5.13...

The "as of" would make clearer that users will need to update the crate to benefit from future Landlock features.

src/lib.rs
```diff
@@ -1,3 +1,75 @@
...
+//! and then inherits some [kernel limitations](https://www.kernel.org/doc/html/latest/userspace-api/landlock.html#current-limitations)
```
gnoack, line 30: (fyi) Good idea to spell this out in the documentation, I'll copy the warning about these two items into the go-landlock documentation as well.

src/lib.rs
```diff
@@ -1,3 +1,75 @@
...
+//! that will be addressed with future kernel releases:
+//! * File reparenting: renaming or linking a file to a different parent directory is always denied.
+//! * Filesystem topology modification: arbitrary mounts are always denied.
+//!
+//! # Compatibility
+//!
+//! This crate is designed to enable security as much as possible on supported systems.
```
gnoack, line 37: I found lines 37-41 difficult to understand; let me suggest the following phrasing (assuming that I understood it correctly):

> This crate is designed to enable the strictest Landlock configuration for the the given kernel on which the program runs.
> The crate will determine compatibility based on the currently running kernel's features,
> not based on the specific rule provided by the caller.
> This way, callers can distinguish between Landlock compatibility issues inherent to the system
> and misconfigured rules (e.g., providing file names that don't exist).

I tried formulating this in active instead of passive voice above, but didn't know the code well enough to be specific; maybe "the crate" would better be "the function Xyz()" or "the class Foo" in some places?

I'm not sure I understood this in its entirety, but it seems that the core point is that callers can distinguish between issues of their own fault (wrong file names?), and issues inherent to the system they run on (missing Landlock support in kernel). Maybe this would be less abstract and easier to explain by having a section that is listing possible error scenarios, for example:

> # Errors
>
> The following errors are possible:
>
> * The kernel does not support the requested Landlock features. In best effort mode, this crate will use the best available security settings (possibly none). Otherwise, the function Foobar() returns error code XYZ
> * The file names used when constructing the rule cannot be opened. This is exposed as PathFdError when bulding a PathFd object using PathFd::new(). (?)

There's a good chance I'm overlooking some parts here :)

src/lib.rs
```diff
@@ -1,3 +1,75 @@
...
+//! When running the same program on different environments,
+//! this helps catch issues with the assumed context
+//! (e.g., whether a directory hierarchy exists or not).
+//!
+//! To make it easier to migrate to a new version of this library,
+//! we use the builder pattern
+//! and designed objects to require the minimal set of method arguments.
+//! Most `enum` are marked as `non_exhaustive` to enable backward-compatible evolutions.
+//!
+//! A simple approach would have been to expose the Landlock [ABI] version
```
gnoack, line 51: I would suggest to phrase this more in terms of "what users are supposed to do", for example:

> In best-effort mode, users of the crate may identify [with the XYZ helper?] which Landlock features are deemed required and which features may be downgraded to use lower security on systems where they can't be enforced. It is discouraged to compare the system's provided Landlock ABI version directly, as it is difficult to track detailed ABI differences, and these will already be handled by the crate's "best effort" mechanism.

NB, I do not remember exactly right now how the difference between 'required' and 'best effort' is expressed in the Rust library for individual features - In my memory this was just a single global toggle?

src/lib.rs
l0kod: Yep, I'll remove "monolith" which is useless anyway.

src/lib.rs
```diff
@@ -1,3 +1,75 @@
...
+//! or required features.
+//! This makes more sense because this directly relies on the access-control semantic.
+//! It makes the library more difficult to implement but simpler to use.
+//! This also makes it easier to audit for most complex requirements
+//! because access rights are explicitly set as required,
+//! instead of relying on manual ABI version mapping.
+//!
+//! ## Test strategy
+//!
+//! Developers should test their sandboxed applications with a compatible kernel
```
gnoack, line 65: "compatible kernel" is vague - they are all "compatible" in some sense, in best effort mode. Maybe it would be better to say more specifically "a kernel that supports all requested Landlock features".

Further up, you are saying "This crate handles the Landlock features available since Linux 5.13" - maybe it would be easiest to refer to that section, as that seems to be the place where you have documented what set of Landlock features are supported by this version of the library?

src/compat.rs
```diff
@@ -4,11 +4,31 @@ use crate::{uapi, Access, AccessError, BitFlags, CompatError};
...
+/// because it could change insidiously over time with a minor crate update
+/// (i.e. non-breaking change with new supported features).
+/// This usage should then be considered indeterministic because requested features
+/// (e.g., access rights)
+/// could not be tied to the application source code.
 #[cfg_attr(test, derive(Debug, PartialEq))]
 #[derive(Copy, Clone)]
 #[non_exhaustive]
 pub enum ABI {
+    /// Kernel not supporting Landlock, either because it is not build with Landlock
```
gnoack, line 27: s/build/built/

src/compat.rs
```diff
@@ -4,11 +4,31 @@ use crate::{uapi, Access, AccessError, BitFlags, CompatError};
 use crate::{make_bitflags, AccessFs};
 
 /// Version of the Landlock [ABI](https://en.wikipedia.org/wiki/Application_binary_interface).
+///
+/// This `enum` may be convenient to get the features supported by a specific Linux kernel,
```
gnoack, line 8: Maybe this comment would benefit by putting the point first and the explanation for it later? For example:

> You can retrieve the supported AccessFs flags for a Landlock ABI version using AccessFs::from_all(ABI::V1).
>
> DO NOT use AccessFs::ALL or BitFlags::<AccessFs>::all() to construct a Landlock rule object -- the full
> set of AccessFs flags is a moving target and would change the semantics of your Landlock rule when
> migrating to a newer version of the Landlock crate.
>
> [longer explanation here]

?

src/compat.rs
```diff
@@ -159,10 +169,66 @@ impl Compatibility {
     }
 }
 
+/// Properly handles runtime unsupported features.
+///
+/// This enables to guarantee consistent behaviors across crate users
```
gnoack, line 174: I'd drop the "enables to", it doesn't add meaning here I think?

> This guarantees consistent behaviors across crate users...

src/compat.rs
```diff
@@ -159,10 +169,66 @@ impl Compatibility {
...
+    /// # Example
+    ///
+    /// Create a ruleset which will at least support execution constraints.
+    ///
+    /// ```
+    /// use landlock::{
+    ///     Access, AccessFs, Compatible, PathBeneath, Ruleset, RulesetCreated, RulesetError, ABI,
+    /// };
+    ///
+    /// fn ruleset_fragile() -> Result<RulesetCreated, RulesetError> {
```
gnoack, line 215: This is an enlightening example! I didn't realize that this is how "best effort" mode can be used feature by feature.

(Slight nit: If I understand correctly, two subsequent .handle_access() calls are possible but should not be needed; does it make more sense to use it with .add_rule() calls?)

src/ruleset.rs
```diff
@@ -98,6 +103,81 @@ fn support_no_new_privs() -> bool {
     )
 }
 
+/// Landlock ruleset builder.
+///
+/// `Ruleset` enables to create a Landlock ruleset in a flexible way
+/// following the builder pattern.
+/// Most build step return a [`Result`] with [`RulesetError`].
```
gnoack, line 110: s/step/steps/

src/lib.rs
```diff
@@ -1,3 +1,75 @@
...
+//!   (e.g., some system or network services).
+//! * Monolith applications dealing with different levels of confidentiality
+//!   (e.g., web browser, email server).
+//!
+//! # Examples
+//!
+//! Examples can be found with the [`Ruleset` documentation](Ruleset)
+//! and the [sandboxer example](https://github.com/landlock-lsm/rust-landlock/blob/master/examples/sandboxer.rs).
+//!
+//! # Current limitations
```
gnoack, line 27: At the level of the "crate" (rather than just Landlock), users may also be surprised to notice that it has enabled `PR_SET_NO_NEW_PRIVS` for them - might be worth documenting somewhere at this level as well?

src/lib.rs
l0kod: I changed with this:
```
//! With [`set_best_effort()`](Compatible::set_best_effort),
//! users of the crate may identify which Landlock features are deemed required
//! and which features may be downgraded to use lower security on systems
//! where they can't be enforced.
//! It is discouraged to compare the system's provided [Landlock ABI](ABI) version directly,
//! as it is difficult to track detailed ABI differences
//! which are handled thanks to the [`Compatible`] trait.
```

src/lib.rs
l0kod: The supported features and their documentation should comes from the ABI enum used, for example `AccessFs::from_all(ABI::V1)`. I need to expand that and link to it.

src/compat.rs
l0kod: Updated code:
```
/// `ABI` enables to get the features supported by a specific Landlock ABI.
/// For example, [`AccessFs::from_all(ABI::V1)`](Access::from_all)
/// gets all the file system access rights defined by the first version.
///
/// Without `ABI`, it would be hazardous to rely on the the full set of access flags
/// (e.g., `BitFlags::<AccessFs>::all()` or `BitFlags::ALL`),
/// a moving target that would change the semantics of your Landlock rule
/// when migrating to a newer version of this crate
/// (i.e. non-breaking change with new supported features).
/// This usage should then be considered indeterministic because requested features
/// (e.g., access rights)
/// could not be tied to the application source code.
///
/// Such `ABI` is also convenient to get the features supported by a specific Linux kernel
/// without relying on the kernel version (which may not be accessible or patched).
```

src/compat.rs
l0kod: I'm working on changes to `handle_access()` that makes it OR multiple calls, which is convenient for this use case and less error-prone in general.

This example may be better with `add_rule()` but it takes more lines. ;)

src/lib.rs
gnoack: Sounds good to me.
(MIcro nit, I'd say "In the default [best effort](Compatible::set_best_effort) mode, ..." to clarify that it's the default)

src/compat.rs
gnoack: Nice, this is much clearer.

(A related but non-documentation question: The method call `AccessFs::from_all(ABI::V1)` reads weird to me when read left to right - is this a Rust idiosyncrasy that refers to the enum 'all' method or attribute? Why not make it more descriptive, for example `AccessFs::as_supported_in(ABI::V1)`?)

l0kod: I first used the `From` trait, which is implemented with the `from()` function to convert types. But I realized that it wasn't wise to only rely on one kind of conversion, hence the `from_all()` which is intended to express *all* defined bits from a specific ABI. In a set of WIP commits, I implemented `from_read()` and `from_write()` which split the bits in two kinds.

The *from_* part seems a bit weird now, but I would like to keep the function names short. The *new_* prefix would be misleading because `from_all()` doesn't create an `AccessFs` but a `BitFlags<AccessFs>`, which I realized is another argument to not use the *from_* prefix… The *as_* prefix has usually a specific meaning in Rust.

What about the `gen_` or `infer_` or `with_` (`AccessFs::with_all()`, `with_read`, `with_write`) prefixes? :thinking:

l0kod: Or maybe `all_from()`, `read_from()`…?

gnoack: I don't have a strong opinion here as I don't know Rust naming idioms well. I'd normally lean on calling methods in a way that the invocation reads like a sentence, but in doubt, idioms like "from()" methods trump that approach.

l0kod: Other idea: `AccessFs::all_flags(ABI::V1) -> BitFlags<AccessFs>`, `read_flags()` and `write_flags()`.

l0kod: Hmm, `read_flags()` is ambiguous because of the verb form.
`get_all_flags()`, `get_read_flags()` and `get_write_flags()` seems better.

l0kod: These `get_*()` function could be colliding with getter methods…
`new_all_flags()`, `new_read_flags()` and `new_write_flags()` seem less ambiguous (even if `new` is usually used to create a type matching the struct type, here `AccessFs`).

l0kod: I added such helpers with #10.

l0kod:
This update should be better.

l0kod:
@gnoack, I know we are discussing about the best-effort approach, but because this PR (and my fixes) predate this discussion, let's first merge it. ;)

l0kod:
I think I took all your review into account, thanks @gnoack! If you spot something else, please feel free to add a comment here or open a new PR.

l0kod:
The documentation is now available here: https://landlock.io/rust-landlock/landlock/

PR #10, opened 2022-08-05
Add helpers

l0kod:
Extend the Access trait with from_read() and from_write(), and add a path_rules() helper.


l0kod:
This changes the documentation, hence the #9 commits.

l0kod:
The `from_read()` and `from_write()` names might not be a good choice though.

PR #11, opened 2022-08-05
Test on CI

l0kod:
It is important to be able to use and test the library without a kernel supporting Landlock.


l0kod:
This PR change the documentation, hence the #9 commits.

l0kod:
This is still a WIP though.

PR #12, opened 2022-08-05
API compatibility improvements

l0kod:
Add several API improvements, including for the compatibility/best-effort handling.

### DIFF
--- a/src/compat.rs
+++ b/src/compat.rs
@@ -125,8 +125,8 @@ lazy_static! {
 }
 
 #[cfg(test)]
-pub(crate) fn can_emulate(mock: ABI, full_support: ABI) -> bool {
-    mock <= *TEST_ABI || full_support <= *TEST_ABI
+pub(crate) fn can_emulate(mock: ABI, partial_support: ABI, full_support: ABI) -> bool {
+    mock < partial_support || mock <= *TEST_ABI || full_support <= *TEST_ABI
 }
 
 #[cfg(test)]
@@ -163,6 +163,8 @@ fn current_kernel_abi() {
 #[cfg_attr(test, derive(Debug))]
 #[derive(Copy, Clone, PartialEq)]
 pub(crate) enum CompatState {
+    /// Initial undefined state.
+    Init,
     /// All requested restrictions are enforced.
     Full,
     /// Some requested restrictions are enforced, following a best-effort approach.
@@ -170,14 +172,15 @@ pub(crate) enum CompatState {
     /// The running system doesn't support Landlock.
     No,
     /// Final unsupported state.
-    Final,
+    Dummy,
 }
 
 impl CompatState {
     fn update(&mut self, other: Self) {
         *self = match (*self, other) {
-            (CompatState::Final, _) => CompatState::Final,
-            (_, CompatState::Final) => CompatState::Final,
+            (CompatState::Init, other) => other,
+            (CompatState::Dummy, _) => CompatState::Dummy,
+            (_, CompatState::Dummy) => CompatState::Dummy,
             (CompatState::No, CompatState::No) => CompatState::No,
             (CompatState::Full, CompatState::Full) => CompatState::Full,
             (_, _) => CompatState::Partial,
@@ -204,11 +207,11 @@ fn compat_state_update_1() {
     state.update(CompatState::No);
     assert_eq!(state, CompatState::Partial);
 
-    state.update(CompatState::Final);
-    assert_eq!(state, CompatState::Final);
+    state.update(CompatState::Dummy);
+    assert_eq!(state, CompatState::Dummy);
 
     state.update(CompatState::Full);
-    assert_eq!(state, CompatState::Final);
+    assert_eq!(state, CompatState::Dummy);
 }
 
 #[test]
@@ -232,10 +235,6 @@ pub struct Compatibility {
     abi: ABI,
     pub(crate) level: CompatLevel,
     pub(crate) state: CompatState,
-    // is_mooted is required to differenciate a kernel not supporting Landlock from an error that
-    // occured with CompatLevel::SoftRequirement.  is_mooted is only changed with update() and only
-    // used to not set no_new_privs in RulesetCreated::restrict_self().
-    is_mooted: bool,
 }
 
 impl From<ABI> for Compatibility {
@@ -244,11 +243,10 @@ impl From<ABI> for Compatibility {
             abi,
             level: CompatLevel::default(),
             state: match abi {
-                // Forces the state as unsupported because all possible types will be useless.
-                ABI::Unsupported => CompatState::Final,
-                _ => CompatState::Full,
+                // Don't forces the state as Dummy because no_new_privs may still be legitimate.
+                ABI::Unsupported => CompatState::No,
+                _ => CompatState::Init,
             },
-            is_mooted: false,
         }
     }
 }
@@ -262,18 +260,10 @@ impl Compatibility {
 
     pub(crate) fn update(&mut self, state: CompatState) {
         self.state.update(state);
-        if state == CompatState::Final {
-            self.abi = ABI::Unsupported;
-            self.is_mooted = true;
-        }
     }
 
     pub(crate) fn abi(&self) -> ABI {
         self.abi
-    }
-
-    pub(crate) fn is_mooted(&self) -> bool {
-        self.is_mooted
     }
 }
 
--- a/src/compat.rs
+++ b/src/compat.rs
@@ -229,7 +229,7 @@ fn compat_state_update_2() {
 #[derive(Clone)]
 // Compatibility is not public outside this crate.
 pub struct Compatibility {
-    pub(crate) abi: ABI,
+    abi: ABI,
     pub(crate) level: CompatLevel,
     pub(crate) state: CompatState,
     // is_mooted is required to differenciate a kernel not supporting Landlock from an error that
@@ -266,6 +266,10 @@ impl Compatibility {
             self.abi = ABI::Unsupported;
             self.is_mooted = true;
         }
+    }
+
+    pub(crate) fn abi(&self) -> ABI {
+        self.abi
     }
 
     pub(crate) fn is_mooted(&self) -> bool {
--- a/src/compat.rs
+++ b/src/compat.rs
@@ -288,7 +288,8 @@ impl Compatibility {
 /// (e.g. applications carefully designed to only be run with a specific set of kernel features),
 /// it may be required to error out if some of these features are not available
 /// and will then not be enforced.
-pub trait Compatible {
+pub trait Compatible: Sized + AsMut<CompatLevel> {
+    // TODO: Update ruleset_handling_renames() with ABI::V3
     /// To enable a best-effort security approach,
     /// Landlock features that are not supported by the running system
     /// are silently ignored by default,
@@ -304,7 +305,7 @@ pub trait Compatible {
     /// Such builders have a set of methods to incrementally build an object.
     /// These build methods rely on kernel features that may not be available at runtime.
     /// The `set_compatibility()` method enables to control the effect of
-    /// the following build method calls starting from this call.
+    /// the following build method calls starting after the `set_compatibility()` call.
     /// Such effect can be:
     /// * to silently ignore unsupported features
     ///   and continue building ([`CompatLevel::BestEffort`]);
@@ -335,36 +336,66 @@ pub trait Compatible {
     /// a [`RulesetCreated`](crate::RulesetCreated) object that inherits the
     /// `Ruleset`'s compatibility configuration.
     ///
-    /// # Example
+    /// # Example with `SoftRequirement`
     ///
-    /// Create a ruleset which will at least support all restrictions provided by
-    /// the [first version of Landlock](ABI::V1), and may also support the
-    /// [`AccessFs::Refer`](crate::AccessFs::Refer) restriction according to the running kernel.
+    /// Let's say an application legitimately needs to rename files between directories.
+    /// Because of [previous Landlock limitations](https://docs.kernel.org/userspace-api/landlock.html#file-renaming-and-linking-abi-2),
+    /// this was forbidden with the [first version of Landlock](ABI::V1),
+    /// but it is now handled starting with the [second version](ABI::V2).
+    /// For this use case, we only want the application to be sandboxed
+    /// if we have the guarantee that it will not break a legitimate usage (i.e. rename files).
+    /// We then create a ruleset which will either support file renaming
+    /// (thanks to [`AccessFs::Refer`](crate::AccessFs::Refer)) or silently do nothing.
     ///
     /// ```
-    /// use landlock::{
-    ///     ABI, Access, AccessFs, CompatLevel, Compatible, PathBeneath, Ruleset, RulesetAttr,
-    ///     RulesetCreated, RulesetError,
-    /// };
+    /// use landlock::*;
     ///
-    /// fn ruleset_fragile() -> Result<RulesetCreated, RulesetError> {
+    /// fn ruleset_handling_renames() -> Result<RulesetCreated, RulesetError> {
     ///     Ok(Ruleset::new()
-    ///         // This ruleset must handle at least all accesses defined by
-    ///         // the first Landlock version (e.g. AccessFs::WriteFile).
-    ///         .set_compatibility(CompatLevel::HardRequirement)
-    ///         // This handle_access() call may now return a wrapped
-    ///         // AccessError<AccessFs>::Incompatible error if Landlock
-    ///         // is not supported by the running kernel.
-    ///         .handle_access(AccessFs::from_all(ABI::V1))?
-    ///         // This ruleset may also handle the AccessFs::Refer right (defined by
-    ///         // the second version of Landlock) if it is supported by the running kernel.
-    ///         .set_compatibility(CompatLevel::BestEffort)
-    ///         // The following handle_access() calls will now never return an error.
+    ///         // This ruleset must either handle the AccessFs::Refer right,
+    ///         // or it must silently ignore the whole sandboxing.
+    ///         .set_compatibility(CompatLevel::SoftRequirement)
     ///         .handle_access(AccessFs::Refer)?
+    ///         // However, this ruleset may also handle other (future) access rights
+    ///         // if they are supported by the running kernel.
+    ///         .set_compatibility(CompatLevel::BestEffort)
+    ///         .handle_access(AccessFs::from_all(ABI::V2))?
     ///         .create()?)
     /// }
     /// ```
-    fn set_compatibility(self, level: CompatLevel) -> Self;
+    ///
+    /// # Example with `HardRequirement`
+    ///
+    /// Security-dedicated applications may want to ensure that
+    /// an untrusted software component is subject to a minimum of restrictions before launching it.
+    /// In this case, we want to create a ruleset which will at least support
+    /// all restrictions provided by the [first version of Landlock](ABI::V1),
+    /// and opportunistically handle restrictions supported by newer kernels.
+    ///
+    /// ```
+    /// use landlock::*;
+    ///
+    /// fn ruleset_fragile() -> Result<RulesetCreated, RulesetError> {
+    ///     Ok(Ruleset::new()
+    ///         // This ruleset must either handle at least all accesses defined by
+    ///         // the first Landlock version (e.g. AccessFs::WriteFile),
+    ///         // or the following handle_access() call must return a wrapped
+    ///         // AccessError<AccessFs>::Incompatible error.
+    ///         .set_compatibility(CompatLevel::HardRequirement)
+    ///         .handle_access(AccessFs::from_all(ABI::V1))?
+    ///         // However, this ruleset may also handle new access rights
+    ///         // (e.g. AccessFs::Refer defined by the second version of Landlock)
+    ///         // if they are supported by the running kernel,
+    ///         // but without returning any error otherwise.
+    ///         .set_compatibility(CompatLevel::BestEffort)
+    ///         .handle_access(AccessFs::from_all(ABI::V2))?
+    ///         .create()?)
+    /// }
+    /// ```
+    fn set_compatibility(mut self, level: CompatLevel) -> Self {
+        *self.as_mut() = level;
+        self
+    }
 
     /// Cf. [`set_compatibility()`](Compatible::set_compatibility()):
     ///
--- a/src/compat.rs
+++ b/src/compat.rs
@@ -160,8 +160,8 @@ fn current_kernel_abi() {
 }
 
 /// Returned by ruleset builder.
-#[cfg_attr(test, derive(Debug, PartialEq))]
-#[derive(Copy, Clone)]
+#[cfg_attr(test, derive(Debug))]
+#[derive(Copy, Clone, PartialEq)]
 pub(crate) enum CompatState {
     /// All requested restrictions are enforced.
     Full,
@@ -174,7 +174,7 @@ pub(crate) enum CompatState {
 }
 
 impl CompatState {
-    pub(crate) fn update(&mut self, other: Self) {
+    fn update(&mut self, other: Self) {
         *self = match (*self, other) {
             (CompatState::Final, _) => CompatState::Final,
             (_, CompatState::Final) => CompatState::Final,
@@ -225,25 +225,30 @@ fn compat_state_update_2() {
     assert_eq!(state, CompatState::Partial);
 }
 
-#[cfg_attr(test, derive(Debug))]
+#[cfg_attr(test, derive(Debug, PartialEq))]
 #[derive(Clone)]
 // Compatibility is not public outside this crate.
 pub struct Compatibility {
     pub(crate) abi: ABI,
-    pub(crate) is_best_effort: bool,
+    pub(crate) level: CompatLevel,
     pub(crate) state: CompatState,
+    // is_mooted is required to differenciate a kernel not supporting Landlock from an error that
+    // occured with CompatLevel::SoftRequirement.  is_mooted is only changed with update() and only
+    // used to not set no_new_privs in RulesetCreated::restrict_self().
+    is_mooted: bool,
 }
 
 impl From<ABI> for Compatibility {
     fn from(abi: ABI) -> Self {
         Compatibility {
             abi,
-            is_best_effort: true,
+            level: CompatLevel::default(),
             state: match abi {
                 // Forces the state as unsupported because all possible types will be useless.
                 ABI::Unsupported => CompatState::Final,
                 _ => CompatState::Full,
             },
+            is_mooted: false,
         }
     }
 }
@@ -251,8 +256,20 @@ impl From<ABI> for Compatibility {
 impl Compatibility {
     // Compatibility is an opaque struct.
     #[allow(clippy::new_without_default)]
-    pub fn new() -> Self {
+    pub(crate) fn new() -> Self {
         ABI::new_current().into()
+    }
+
+    pub(crate) fn update(&mut self, state: CompatState) {
+        self.state.update(state);
+        if state == CompatState::Final {
+            self.abi = ABI::Unsupported;
+            self.is_mooted = true;
+        }
+    }
+
+    pub(crate) fn is_mooted(&self) -> bool {
+        self.is_mooted
     }
 }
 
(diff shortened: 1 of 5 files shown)


l0kod:
@gnoack, commit [compat: Switch from set_best_effort() to set_compatibility()](https://github.com/landlock-lsm/rust-landlock/pull/12/commits/51a6967cdda19164a5e9b8405bebc07b6a9e32d5) should particularly be of interest. I developed that in November (before the Landlock's refer right was developed) though.

gnoack:
> @gnoack, commit [compat: Switch from set_best_effort() to set_compatibility()](https://github.com/landlock-lsm/rust-landlock/pull/12/commits/51a6967cdda19164a5e9b8405bebc07b6a9e32d5) should particularly be of interest. I developed that in November (before the Landlock's refer right was developed) though.

Thanks, I'll have a look. (might take me a bit longer to re-familiarize myself with the Rust code)

src/compat.rs
```diff
@@ -221,17 +245,36 @@ pub trait Compatible {
...
+    /// Never returns a compatibility error.
+    BestEffort,
+    /// Takes into account the requests if they are supported by the running system,
+    /// or silently ignores all the following requests otherwise.
+    /// Never returns a compatibility error.
+    /// If not supported,
+    /// the call to [`RulesetCreated::restrict_self()`](crate::RulesetCreated::restrict_self())
+    /// will return a
+    /// [`RestrictionStatus { ruleset: RulesetStatus::NotEnforced, no_new_privs: false, }`](crate::RestrictionStatus).
+    SoftRequirement,
```
gnoack, line 277: I tried to figure out what the `SoftRequirement` does, but am not sure I fully understood it.

It has taken me a while to make sense of the documentation (and I had a hard time following the indirection through the different types and traits, due to my unfamiliarity with Rust, I think).

Do I understand correctly that:

* When using `HardRequirement`, you'll get an error straight away when adding the specific file hierarchy rule, if the given access rights are not supported by the kernel.
* When using `SoftRequirement`, the compatibility error will be "postponed" and only be returned by the call to `RulesetCreated::restrict_self()`.
  * Both in "best effort" and "soft" mode, you cannot get a compatibility error directly. 
* I suspect (but am not fully sure) that in `SoftRequirement` mode, if `restrict_self()` returns this error, nothing was enforced.
* Is it right that the compatibility mode takes effect both on the Ruleset when adding handled accesses as well as on the CreatedRuleset when adding rules?
  * I only noticed that this is meant to apply to handled accesses and be stateful when I looked at tests in ruleset.rs.
  * Is it right that the term "request" in this documentation is the abstraction over both handled accesses as well as individual rules? (I wonder whether this generality is worth it. Why not make it a flag in the `add_rule()` and `handle_access()` functions? This would remove this additional "request" concept and the statefulness.)

As you can see, I'm somewhat confused what this means. If some of what I'm saying here doesn't make sense, a good chunk of the reason is probably that I don't understand Rust well. %-) so some of it is probably pointless.

P.S.: I tried following the https://docs.rs/landlock/ link from the README to refresh my memory, and found the documentation page to be empty. I'm not sure what went wrong there?

l0kod: > Do I understand correctly that:
> 
> * When using `HardRequirement`, you'll get an error straight away when adding the specific file hierarchy rule, if the given access rights are not supported by the kernel.

Right, it as an impact on the returned `Result<>` type, which is then automatically caught with the trailing `?`.

> * When using `SoftRequirement`, the compatibility error will be "postponed" and only be returned by the call to `RulesetCreated::restrict_self()`.

Right, it moot the ruleset, but always return `Ok(_)` (unwrapped with `?`).

>   * Both in "best effort" and "soft" mode, you cannot get a compatibility error directly.

Right, but you can still get the status (`RestrictionStatus`) of the enforced (or not) ruleset at the end.

> * I suspect (but am not fully sure) that in `SoftRequirement` mode, if `restrict_self()` returns this error, nothing was enforced.

Right, it is especially useful e.g. for kernel not supporting the "refer" right: in this case it would not return an error when building the ruleset but it will not call landlock_restrict_self(), only return a status (`RestrictionStatus`).

> * Is it right that the compatibility mode takes effect both on the Ruleset when adding handled accesses as well as on the CreatedRuleset when adding rules?

Yes, the compatibility level is taken into account for all `handle_access()` and `add_rule()` calls thanks to the `impl Compatible for Ruleset` and `impl Compatible for RulesetCreated`.

>   * I only noticed that this is meant to apply to handled accesses and be stateful when I looked at tests in ruleset.rs.

`Ruleset` and `RulesetCreated` are indeed stateful, which enable them to check if access rights are consistent.

>   * Is it right that the term "request" in this documentation is the abstraction over both handled accesses as well as individual rules? (I wonder whether this generality is worth it. Why not make it a flag in the `add_rule()` and `handle_access()` functions? This would remove this additional "request" concept and the statefulness.)

The requested restrictions are the access rights passed to `handled_access()` and `add_rule()`. They are what the developer would like to enforce if possible (according to `CompatLevel`). We could replace the `set_compatibility()` call with a dedicated argument in `handle_access()` and `add_rule()` (or `PathBeneath::new()` and other future type of rules), but that increases complexity for developers whereas most of the time they want a best-effort approach. This approach is simpler for most use cases (i.e. no call to `set_compatibility()`), but still enables to fine-tune the Landlock compatibility. I think the builder pattern is cleaner and all optional arguments should be optional. ;)

> As you can see, I'm somewhat confused about a few things here. If some of what I'm saying here doesn't make sense, a good chunk of the reason is probably that I don't understand Rust that well, so some of it is probably pointless and might not be actionable. I'm sending you my vague understanding anyway, in the hope that it'll help to surface some problems that a Rust beginner might run into. %-)

Your feedback is useful. My use of Rust may be a bit complex for this library internals but I'm looking for a simpler but still flexible and updatable-without-breaking-the-strong-type-system developer API.

> P.S.: I tried following the https://docs.rs/landlock/ link from the README to refresh my memory, and found the documentation page to be empty. I'm not sure what went wrong there?

Right, I didn't uploaded to crates.io (which builds the doc and exposes it on docs.rs) yet. You can generate the documentation with `cargo doc` and I just published the current documentation at https://landlock.io/rust-landlock/landlock/ (a lot of Landlock nesting but ¯\\\_(シ)\_/¯ ).

gnoack: > > * I suspect (but am not fully sure) that in `SoftRequirement` mode, if `restrict_self()` returns this error, nothing was enforced.
> 
> Right, it is especially useful e.g. for kernel not supporting the "refer" right: in this case it would not return an error when building the ruleset but it will not call landlock_restrict_self(), only return a status (`RestrictionStatus`).

So, to make sure I understand; you'd use it like this (pseudocode)?

* set compatibility level to "hard requirement"
* add handled access for all the V1 rights
* set compatibility level to "soft requirement"
* add handled access for "refer"
* create()
* add_rule("/tmp", read+write+...+refer)
* add_rule("/usr", read+readdir)
* restrict_self()

and when that is running on a kernel with only V1 support, then it would raise a less dramatic error during `restrict_self()`, but it would still not restrict anything? And the intent is to simplify error handling?

Some small questions:

* Does there need to be another `set_compatibility()` call between `create()` and `add_rule()`?
* Which would be the case where you would still use the hard requirement? If the only difference is in the time of surfacing the error, why not just always surface it only during `restrict_self()`?

> > * Is it right that the term "request" in this documentation is the abstraction over both handled accesses as well as individual rules? (I wonder whether this generality is worth it. Why not make it a flag in the `add_rule()` and `handle_access()` functions? This would remove this additional "request" concept and the statefulness.)
> 
> The requested restrictions are the access rights passed to `handled_access()` and `add_rule()`. They are what the developer would like to enforce if possible (according to `CompatLevel`). We could replace the `set_compatibility()` call with a dedicated argument in `handle_access()` and `add_rule()` (or `PathBeneath::new()` and other future type of rules), but that increases complexity for developers whereas most of the time they want a best-effort approach. This approach is simpler for most use cases (i.e. no call to `set_compatibility()`), but still enables to fine-tune the Landlock compatibility. I think the builder pattern is cleaner and all optional arguments should be optional. ;)

OK then :)

l0kod: > > > * I suspect (but am not fully sure) that in `SoftRequirement` mode, if `restrict_self()` returns this error, nothing was enforced.
> > 
> > 
> > Right, it is especially useful e.g. for kernel not supporting the "refer" right: in this case it would not return an error when building the ruleset but it will not call landlock_restrict_self(), only return a status (`RestrictionStatus`).
> 
> So, to make sure I understand; you'd use it like this (pseudocode)?
> 
> * set compatibility level to "hard requirement"
> * add handled access for all the V1 rights
> * set compatibility level to "soft requirement"
> * add handled access for "refer"
> * create()
> * add_rule("/tmp", read+write+...+refer)
> * add_rule("/usr", read+readdir)
> * restrict_self()
> 
> and when that is running on a kernel with only V1 support, then it would raise a less dramatic error during `restrict_self()`, but it would still not restrict anything? And the intent is to simplify error handling?

Correct! In fact I was waiting for a v2 to test this behavior. ;)
I just added support for ABI v1 (no PR yet) and here is your test (which passes):

```rust
#[test]
fn ignore_abi_v2_with_abi_v1() {
    assert_eq!(
        Ruleset::from(ABI::V1)
            .set_compatibility(CompatLevel::HardRequirement)
            .handle_access(AccessFs::from_all(ABI::V1))
            .unwrap()
            .set_compatibility(CompatLevel::SoftRequirement)
            .handle_access(AccessFs::Refer)
            .unwrap()
            .create()
            .unwrap()
            .add_rule(PathBeneath::new(
                PathFd::new("/tmp").unwrap(),
                AccessFs::from_all(ABI::V2)
            ))
            .unwrap()
            .add_rule(PathBeneath::new(
                PathFd::new("/usr").unwrap(),
                make_bitflags!(AccessFs::{ReadFile | ReadDir})
            ))
            .unwrap()
            .restrict_self()
            .unwrap(),
        RestrictionStatus {
            ruleset: RulesetStatus::NotEnforced,
            no_new_privs: false,
        }
    );
}
```

It's verbose but it is for the test. #10 and the current features makes writing such rule much more easier and less verbose. I guess I should create the PR for ABI v2 and rebase this one on top of it with your test.

Without the any `set_compatibility()` call, this example returns `RestrictionStatus { ruleset: PartiallyEnforced, no_new_privs: true }`

I noticed there is an issue with the HardRequirement when added after the first rule though, I'll fix that.

> Some small questions:
> 
> * Does there need to be another `set_compatibility()` call between `create()` and `add_rule()`?

You only call `set_compatibility()` when you want the following part of the builder behavior to change.

> * Which would be the case where you would still use the hard requirement? If the only difference is in the time of surfacing the error, why not just always surface it only during `restrict_self()`?

The use case for the hard requirement is when you control the kernel and the sandbox, and you want to make sure that you launch a process in a sandboxed environment.

So yes, the time of surfacing the error changes. The use case for the soft requirement is to not apply the sandbox when a specific access right (e.g. refer) is not supported, without throwing an error because you still want to sandbox as much as possible, but want to know e.g. if a legitimate path is not accessible whereas it should be, which brings confidence in the sandbox configuration even when the running kernel doesn't (fully) support it. I want to help developers catch configuration errors as soon as possible, e.g. in alpha releases.

The difference is really in the way Rust helps to handle errors (i.e. with the `?`) and the error (strong) types.

gnoack: Ah thanks, that makes it clearer; the example really helped me understand this better.

src/ruleset.rs
```diff
@@ -280,13 +250,13 @@ impl Ruleset {
...
-            match self.compat.abi {
+            match self.compat.abi() {
                 ABI::Unsupported => {
                     #[cfg(test)]
-                    assert_eq!(self.compat.state, CompatState::Final);
+                    assert_eq!(self.compat.state(), CompatState::Final);
                     Ok(RulesetCreated::new(self, -1))
                 }
-                ABI::V1 => {
+                ABI::V1 | ABI::V2 => {
```
dburgener, line 259: It looks like you have this pattern 3 times in this file around a uapi call. I guess the idea is that more explicitness will be less error prone in the event that a syscall changes in the future?  My worry is that this list will grow long over a lot of ABI versions and could become more error prone that way.  If just matching `_` isn't amenable to you, maybe derive PartialOrd on ABI or something and match a range of versions instead of a full list?

I guess you understand better than me how this api might conceivably evolve in the future, so maybe I'm missing something here?

src/lib.rs
```diff
@@ -76,78 +77,166 @@ extern crate enumflags2;
...
+                if abi < partial && error_if_abi_lt_partial {
+                    // TODO: Check exact error type; this may require better error types.
+                    assert!(matches!(ret, Err(TestRulesetError::Ruleset(_))));
+                } else {
+                    assert!(matches!(
+                        ret,
+                        Ok(RestrictionStatus {
+                            ruleset: ruleset_status,
+                            no_new_privs: true
+                        }) if ruleset_status == if abi >= full {
```
dburgener, line 139: Maybe it's just me, but I find nesting this if inside the comparison inside a match to be pretty annoying to read and reason about.  I think the code would be more readable if the whole "if abi >= full" clause with its branches was pulled out before the assert, to keep the arguments put into the nested macro calls more simple and straightforward.

src/compat.rs
dburgener: >  The use case for the soft requirement is to not apply the sandbox when a specific access right (e.g. refer) is not supported, without throwing an error because you still want to sandbox as much as possible, but want to know e.g. if a legitimate path is not accessible whereas it should be, which brings confidence in the sandbox configuration even when the running kernel doesn't (fully) support it. I want to help developers catch configuration errors as soon as possible, e.g. in alpha releases.

How does this use case differ from BestEffort?

l0kod: > How does this use case differ from BestEffort?

A process will not be restricted at all with `SoftRequirement` if all following requested features are not supported, whereas `BestEffort` will restrict it as much as possible. One use case for `SoftRequirement` is when we require a set of Landlock features (e.g. the ability to rename a file, which is denied by kernels only supporting Landlock v1) and we should not sandbox the process if the running kernel doesn't support it because it would restrict too much the process and cause issues.

src/ruleset.rs
l0kod: You're right, this notation will not scale. Because there is no specific handling of the Landlock syscalls according to the Landlock ABI versions for now, I will use `_`. Implementing and using a range for `ABI` will be useful later though.

src/lib.rs
l0kod: Indeed, I'll move this one and the other `if … == if …` out of the `matches!()` call.

src/compat.rs
dburgener: Thanks, that clears things up a lot.  Is there a functional difference between SoftRequirement and HardRequirement with the caller discarding the error?

I think the thing I keep getting stuck on with this is I'm not sure how "don't give me an error" is ever a sensible thing when the caller could simply observe the error and move on.  If the desired end state is "don't enforce if not supported", I think I might want to at least log that I'm not enforcing, which would mean receiving the error, but treating it as non-fatal.  If I can treat the error as non-fatal in caller code, then I don't see why I would prefer HardRequirement.  Maybe I'm missing something though?

l0kod: > Thanks, that clears things up a lot. Is there a functional difference between SoftRequirement and HardRequirement with the caller discarding the error?

No, that should be the same, but an error can occur at most method calls with `HardRequirement` (e.g. when adding a new rule, handling a new access right…), whereas with `SoftRequirement` the "error" is only a status error (final `RestrictionStatus`), not a Rust error (i.e. `?`/`Result`).

> I think the thing I keep getting stuck on with this is I'm not sure how "don't give me an error" is ever a sensible thing when the caller could simply observe the error and move on. If the desired end state is "don't enforce if not supported", I think I might want to at least log that I'm not enforcing, which would mean receiving the error, but treating it as non-fatal. If I can treat the error as non-fatal in caller code, then I don't see why I would prefer HardRequirement. Maybe I'm missing something though?

The main idea is to catch *unexpected errors*, for instance a file path not accessible. The goal is to have guarantees that the sandboxing specific syscalls (e.g. not the required open syscalls) are ignored when not available but all the require actions needed by these syscalls are not buggy, even on a system not supporting Landlock. Once Landlock is supported by such systems, there is only a few possible errors that can happen and these errors should already be caught by supported systems (dev workstation or beta releases); only the user/host-specific configuration should be an issue but it was already checked before that.

If the API encourages developers to ignore errors, whether they are legitimate or not, then these errors would be useless because ignored by everyone.

The end result of a ruleset creation, if there is no unexpected errors, is stored in the final `RestrictionStatus`. This can and should be used to log that a system doesn't support a set of Landlock features and the status of the applied restrictions.

dburgener: So it's fundamentally ergonomic?  In other words, one could use HardRequirement instead, match on the result, proceed to call restrict_self() on Ok(_), and then handle error codes as decided by the application (ie silently ignore an error about the ABI not supporting the check and not sandbox, while taking some other action on "unexpected" errors).

l0kod: That would work if the `HardRequirement` is for the whole ruleset building, but the `set_compatibilty()` enables to identify different part of the ruleset as (really) required or not. If it is not a `HardRequirement`, the next steps of the ruleset building go on and we can check as much as possible the sandbox configuration (e.g. make sure all app configured paths exist and are indeed accessible directories…), but without enforcing the partial ruleset even if all these next steps are OK.

l0kod: Among the checks that are useful, there are the rule access right consistency according to the handled access rights, file paths types, and any other consistency current and *future consistency checks* that may be done by the kernel. In a way, this library emulates the consistency checks performed by the kernel to make sure that once users are using a kernel supporting Landlock, there is a good chance that the sandboxing will work as expected.

dburgener: Ah, that makes sense, I think.  So there's a real tangible benefit in delaying the error by allowing the remaining steps to proceed as though there had been no error and potentially catching issues there.  And I haven't really thought through all the switching back and forth scenarios.

So I guess this feature makes sense to have.  It definitely needs better documentation, particularly including examples and use cases.  But maybe with substantial clarification it will make sense.

l0kod:
I still need to update the documentation.

l0kod:
@cd-work, @andreaphylum, if you can take a look that would be nice.

l0kod:
@gnoack and @dburgener, I updated the documentation, added tests, and cleaned up some code to make it more generic. I removed unrelated patches that are either already merged or will be part of an other PR.

src/compat.rs
cd-work: Yeah just to jump on I had pretty similar questions to @gnoack when originally looking at the documentation. It's a very difficult concept to grasp for me.

I personally couldn't see a usecase where I would use this over just handling the errors for `add_rule` myself, but maybe that's just because I'm still not completely grasping the intended goal of the feature.

l0kod: Because we now have the `RulesetAttr` trait, we can handle ruleset builds by reference, which may be useful to manage error per build method call, but that is not the case when building a ruleset in one go.

One example is when an application requires at least the current kernel to be able to move files between directories. For this use case, Landlock must support the `Refer` access right. However, if the `Refer` access right is not handled, it will brake the application. An easy way to handle this case is to set `CompatLevel::SoftRequirement` for the `Refer` access right, which will silently ignore the whole ruleset if the kernel doesn't support this, which is what the app developer wants. It is a way to programmatically identify the minimum requirement without boilerplate error management.

l0kod: @cd-work, I added a new example (one for `SoftRequirement` and another for `HardRequirement`) and improve the related explanations.

cd-work: > It is a way to programmatically identify the minimum requirement without boilerplate error management.

Maybe it's just a case of having different requirements for the things I'd use landlock for, but to me it seems more like a gun to shoot yourself in the foot with rather than something that makes error management easier. I feel like sandboxing failure is one of those scenarios where I'd want to be **really** explicit about handling errors.

Those are just my 2 cents though, I'm sure that the addition of this feature was motivated by some relevant requirement.

The new examples certainly make things more clear. Having it laid out there in a single builder with exact documentation on what the result will actually do is very helpful.

One thing that might be slightly confusing is how `SoftRequirement` and `HardRequirement` play together. Intuitively I'd say that `HardRequirement` should always have priority and based on the examples it seems like that is the case. So that way you'd always be on the safe side at least.

l0kod: > Maybe it's just a case of having different requirements for the things I'd use landlock for, but to me it seems more like a gun to shoot yourself in the foot with rather than something that makes error management easier.

I'd like to better understand your point here. Why do you think this API could be dangerous?

> I feel like sandboxing failure is one of those scenarios where I'd want to be **really** explicit about handling errors.

There is mainly two groups of users for sandboxing:
* Developers of security-focused applications (e.g. sandboxers like bubblewrap) who want to be really explicit about handling errors and probably not launch a third-party application if it is not sandboxed. Another use case is for security-tailored distros/systems/appliances that know that the running kernel must support a specific Landlock version or it should generate an error (for tests and production). In these cases they should use the `HardRequirement` compatibility level.
* Developers of non-security-focused applications who want to follow a best-effort approach for sandboxing when the required sandbox features are available, but who want to launch their application anyway. In this case they should use the `BestEffort` *and* the `SoftRequirement` compatibility level.

This PR really just adds the `SoftRequirement` semantic.

### Why a `SoftRequirement` compatibilty level?

The main motivation is for the `AccessFs::Refer` right. Indeed, some access rights are forbidden by default even if not handled by a ruleset. That is the case for the "refer" access right for Landlock v1 because this cannot be explicitly handled. See @gnoack's document [Landlock File System Access Model](https://docs.google.com/document/d/1SkFpl_Xxyl4E6G2uYIlzL0gY2PFo-Nl8ikblLvnpvlU/edit) and especially the *Example: Landlock V2* section.

If an application needs the `AccessFs::Refer` right to properly work (e.g. because it needs to rename files between different directories), we cannot sandbox the application with kernel only supporting Landlock v1, but only ABI >= v2.

I think it makes more sense to not require library users to know this kind of "side effects" of a particular Landlock ABI. However, it makes sense for users to specify which access to each resources (i.e. file path) are legitimately required, either because the application cannot properly work without them (hence the `SoftRequirement`) or because it might be a good idea to opportunistically add more restrictions (hence the `BestEffort`). This library embed the Landlock ABI specific knowledge with the compatibility mechanism.

There are other ways to implement this kind of mechanism, but I think that the `set_compatibility()` method fits well with the builder pattern: it is an optional call and it is a single interface to change (parts of) the build object behavior.

l0kod: Relevant thread: [Landlock best-effort](https://lore.kernel.org/landlock/7290b021-c90f-76e9-7868-ed612595b411@digikod.net/)

l0kod:
I revamped a big part of the compatibility code (mostly with the latest commit) to make it safer and give more compatibility guarantees. The user API is unchanged.

l0kod:
For now, some test code fail to run with Ubuntu 18 but the revamped code is working as expected. I'll fix that soon.

src/compat.rs
cd-work: > I'd like to better understand your point here. Why do you think this API could be dangerous?

It's been a while due to the holidays, but here goes:

I think the "footgun" issue here is mainly that you need to **really** understand what it is doing, because otherwise it will do something less restrictive than what you'd expect. One might for example think that adding `AccessFs::Refer` as a soft requirement will just ignore the `AccessFs::Refer` bit and you're all good for the rest of the sandbox. But the fact that it drops the sandbox entirely opens up people to unwelcome surprises.

This is especially relevant for people that are from category 1 (bubblewrap etc.), which might look at this initially and think "sweet, I can just use soft requirements and then it'll always allow Refer and just enforce it in version 2", whereas what actually happens is that it allows everything.

l0kod: > I think the "footgun" issue here is mainly that you need to **really** understand what it is doing, because otherwise it will do something less restrictive than what you'd expect. One might for example think that adding `AccessFs::Refer` as a soft requirement will just ignore the `AccessFs::Refer` bit and you're all good for the rest of the sandbox. But the fact that it drops the sandbox entirely opens up people to unwelcome surprises.
>
> This is especially relevant for people that are from category 1 (bubblewrap etc.), which might look at this initially and think "sweet, I can just use soft requirements and then it'll always allow Refer and just enforce it in version 2", whereas what actually happens is that it allows everything.

Hmm, this is the description of `BestEffort`. Would renaming `SoftRequirement` to something else help? How would you improve the situation?

I get that one issue would be that users forget to "reset" the compatibility level by calling `set_compatibility(CompatLevel::BestEffort)` after switching to `SoftRequirement` for `AccessFs::Refer`. Instead of changing the ruleset's compatibility level, what about tuning the access types? That could be something like:
```rust
ruleset.handle_access(AccessFs::Refer.set_compatibility(CompatLevel::SoftRequirement))?;
```
This may look a bit verbose but it would be easy to use some helpers to  tune a set of access rights, and this use case is not the "default safe-for-most-sandboxing-use-cases" one.

cd-work: Changing the name to something more alarming was something I did consider myself as well, but I wasn't able to come up with a good suggestion.

I'd go for something akin to `CompatLevel::NoSandboxIfMissing`, but that doesn't exactly roll off the tongue and fit in well with the rest. But it would be a lot more descriptive and I think something *like* this could communicate better what the actual behavior is.

And while I think that not making it sticky would probably be a good idea too, because people tend to forget stuff a lot, I don't think that would address my concerns completely.

cd-work: Yeah I like that, especially the "disable_sandbox_if_unmet" name is **very** clear and I don't see anyone shooting themselves in the foot with that. There might be a shorter option that says the same, but I like its clarity.

bjorn3: I agree. This proposal is clearer and much easier to write IMHO.

l0kod:
I gave a talk about the challenges of backward and forward compatibility that this PR tries to address: https://fosdem.org/2023/schedule/event/rust_backward_and_forward_compatibility_for_security_features/

l0kod:
> I gave a talk about the challenges of backward and forward compatibility that this PR tries to address: https://fosdem.org/2023/schedule/event/rust_backward_and_forward_compatibility_for_security_features/

The video of the talk is now online.

mrcnski:
Exciting update :) Unfortunately I've tried this branch and it seems to break our tests. Essentially we have some simple code like this:

```rs
let fs_exceptions = path_beneath_rules(
    &[path1],
    AccessFs::from_read(V1),
);
let status = Ruleset::new()
    .handle_access(AccessFs::from_all(V1))?
    .create()?
    .add_rules(fs_exceptions)?
    .restrict_self()?;
```


Previously this resulted in `FullyEnforced` but now gives `PartiallyEnforced`. Is this expected? What changed in the compatibility logic?

l0kod:
> Previously this resulted in `FullyEnforced` but now gives `PartiallyEnforced`. Is this expected? What changed in the compatibility logic?

Thanks for testing. The behavior should not change, but I cannot reproduce your test. The `check_ruleset_support` helper should test all meaningful combinations, and it didn't change with these commits.

Can you give more context? What is the kernel version you're using? Does `path1` exists? Can you bisect to identify the faulty commit?

mrcnski:
> Thanks for testing. The behavior should not change, but I cannot reproduce your test. The `check_ruleset_support` helper should test all meaningful combinations, and it didn't change with these commits.

Weird. I double-checked to be sure, and my tests definitely pass with `landlock 0.2.0` and fail with this branch. No other changes to my code.

> What is the kernel version you're using?

```sh
$ uname -r
5.19.0-1-amd64
```

> Does path1 exists?

Yep. The full code for the tests is here: https://github.com/paritytech/polkadot/pull/7580/files#diff-83d70e8d788d8ad4e6e8b83feb6afa9e9fdc4ccf1d39584b0f2a663af943ed38R168.

> Can you bisect to identify the faulty commit?

I'll try. 👍

mrcnski:
Result of bisect:

```
38d15828845d8945fc950f51cc3723abdf2c399c is the first bad commit
```

l0kod:
I tried to reproduce your code with this patch on Linux 5.19.17 but it works before and after this PR:
```diff
diff --git a/src/lib.rs b/src/lib.rs
index fe62032c5899..b2fcf5b94364 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -203,6 +203,21 @@ mod tests {
         );
     }

+    #[test]
+    fn ruleset_compat_v1() {
+        let status = Ruleset::new()
+            .handle_access(AccessFs::from_all(ABI::V1))
+            .unwrap()
+            .create()
+            .unwrap()
+            .add_rules(path_beneath_rules(["/"], AccessFs::from_read(ABI::V1)))
+            .unwrap()
+            .restrict_self()
+            .unwrap();
+
+        assert_eq!(status.ruleset, RulesetStatus::FullyEnforced);
+    }
+
     #[test]
     fn ruleset_enforced() {
         let abi = ABI::V1;
```

Could you please try this patch or try to write a test that works before this PR but not after?

mrcnski:
Yeah, that passes, but this pared-down version of our test fails:

```rs
#[test]
fn ruleset_compat_v1_temp_file() {
    const TEXT: &str = "foo";
    let tmpfile = tempfile::NamedTempFile::new().unwrap();
    let path = tmpfile.path();

    // Confirm the file exists and is write-able and read-able.
    std::fs::write(path, TEXT).unwrap();
    let s = std::fs::read_to_string(path).unwrap();
    assert_eq!(s, TEXT);

    let status = Ruleset::new()
        .handle_access(AccessFs::from_all(ABI::V1))
        .unwrap()
        .create()
        .unwrap()
        .add_rules(path_beneath_rules([path], AccessFs::from_read(ABI::V1)))
        .unwrap()
        .restrict_self()
        .unwrap();

    assert_eq!(status.ruleset, RulesetStatus::FullyEnforced);
}
```

This uses `tempfile = "3.3.0"`. I tried also with a regular file instead of a temp file and it also fails:

```rs
#[test]
fn ruleset_compat_v1() {
    let path = std::path::Path::new("Cargo.toml");

    // Confirm the file exists and is read-able.
    let s = std::fs::read_to_string(path).unwrap();
    assert!(!s.is_empty());

    let status = Ruleset::new()
        .handle_access(AccessFs::from_all(ABI::V1))
        .unwrap()
        .create()
        .unwrap()
        .add_rules(path_beneath_rules([path], AccessFs::from_read(ABI::V1)))
        .unwrap()
        .restrict_self()
        .unwrap();

    assert_eq!(status.ruleset, RulesetStatus::FullyEnforced);
}
```

l0kod:
Ok, this is indeed the result of a fix. This issue was missing checks for access rights of rules added with `add_rules()` according to the file type. Indeed, allowing all access rights on a file will make the kernel return an error because some of them don't make sense for a file (e.g. create a new file... in a file). The commit you pointed out fixes this behavior, which resulted of a visible behavior change for your code.

I changed a few things:
- added the `AccessFs::from_file()` helper;
- changed `path_beneath_rules()` to automatically check and downgrade access rights for files (i.e. remove superfluous directory-only access rights);
- added tests, especially the (full) `too_much_access_rights_for_a_file()` and `path_beneath_rules_with_too_much_access_rights_for_a_file()` (more generic than the one we wrote).

This should fix your test and keep the same behavior for your (legitimate) use case.

mrcnski:
Thanks, it works now!

l0kod:
I plan to merge this next week if there is no issue found.

Cc @mrcnski @bjorn3 @cd-work @int5-grey

PR #16, opened 2022-08-26
Take mutable reference in ruleset functions

cd-work:
This patch changes the functions on the `RulesetCreated` struct to take
`&mut self` instead of moving `self`.

While taking `self` is extremely useful for `restrict_self` to ensure
correct API usage, it can get very restrictive on the `add_rule(s)` /
`set_no_new_privs` methods. When storing the `RulesetCreated` in your
own struct for example, you'd have to make use of `Option` or move
`self` yourself, making the design somewhat "viral".

The following codeblock is made possible through this patch:

```rs
struct MySandbox {
    ruleset: RulesetCreated,
}

impl MySandbox {
    fn new() -> Self {
        Self { ruleset: Ruleset::new().create().unwrap() }
    }

    fn add_rule(&mut self, rule: MyRuleFormat) {
        self.ruleset.add_rule(rule.into())?;
    }

    fn restrict_self(self) {
        self.ruleset.restrict_self().unwrap();
    }
}
```

I think this change improves the APIs flexibility while not
significantly decreasing legibility for applications that do not require
this flexibility.

This patch also restructures the `add_rule` function a little bit to
hopefully help with legibility.

---

First of all I'd like to say thank you for going out of your way to offer a Rust library for this. It's great to see officially supported Libraries like this especially in a security-sensitive context.

I'm not sure if these changes are desired, or if patches are desired at all at this point, but I'm currently playing around with an implementation that abstracts over Landlock and ran into this problem immediately. Feel free to just close this PR if you have no interest in these changes.


l0kod:
Thanks!

> While taking `self` is extremely useful for `restrict_self` to ensure
> correct API usage,

Indeed, that is the original intent.

> it can get very restrictive on the `add_rule(s)` /
> `set_no_new_privs` methods. When storing the `RulesetCreated` in your
> own struct for example, you'd have to make use of `Option` or move
> `self` yourself, making the design somewhat "viral".

I agree that's unfortunate. I wanted to make it possible to chain build methods, which the current design enables, but it is indeed not perfect.

I guess the `Ruleset` consuming builder should not be an issue though.

About the `RulesetCreated` builder, I didn't found a way to use non-consuming and consuming parts with the same type at the time, but I now think it might be possible to have a more ergonomic builder using something like this:
```rust
trait RulesetCreatedRef where Self: Sized {
    fn get_mut(&mut self) -> &mut Builder; // Should be private somehow.
    fn add_rule(mut self, …) -> Result<Self, RulesetError> {}
}

impl RulesetCreated {
    fn restrict_self(self) -> Result<RestrictionStatus, RulesetError> {}
}

impl RulesetCreatedRef for RulesetCreated {}
impl RulesetCreatedRef for &mut RulesetCreated {}
```

I just quickly tested such approach and it now seems much more usable thanks to the non-lexical lifetime improvements. What do you think?

Anyway, I need to merge #12 before that (hopefuly this week), so this commit will require a rebase.

> I'm not sure if these changes are desired, or if patches are desired at all at this point, but I'm currently playing around with an implementation that abstracts over Landlock and ran into this problem immediately. Feel free to just close this PR if you have no interest in these changes.

I'm looking for constructive feedback like yours! Out of curiosity, what is the abstraction around Landlock?

cd-work:
> I just quickly tested such approach and it now seems much more usable thanks to the non-lexical lifetime improvements. What do you think?

This might get a bit boilerplatey, implementing the same thing for the same thing twice basically. I'm not sure I've seen this approach before, so I don't know how well it would work. I could definitely see it run into some issues, but would have to try it out to confirm.

> Anyway, I need to merge #12 before that (hopefuly this week), so this commit will require a rebase.

:+1: 

> I'm looking for constructive feedback like yours! Out of curiosity, what is the abstraction around Landlock?

Currently it's not public yet, but the abstraction is basically just to get an OS-independent sandboxing API. So on Linux it's fundamentally still just landlock, though that might change based on required features since landlock is still just doing Filesystem (for now).

l0kod:
> > Gaol might be interesting in this respect but unfortunately it seems to not be maintained anymore.
> 
> That's actually what I've tried at first. But the Linux implementation is just broken missing the `open` and `openat` syscalls in their seccomp profile. And even after fixing that I still would have had to bring everything up to date with Rust and implement write access. So at that point it was just easier starting from scratch with landlock considering it makes the most difficult part of sandboxing in gaol (the fs) so trivial.

I think it's worth looking at this fork: https://github.com/Pernosco/gaol/tree/downstream

cd-work:
This PR got automatically closed when I deleted my personal fork, but I don't think the suggestions made in this PR are the best way forward anyway, so any potential improvements should probably go into separate PRs.

PR #20, opened 2022-10-28
Non-consuming RulesetCreated alternative

l0kod:
This is an alternative implementation of #19, as I described in #16.

I think it is simpler than #19, with no duplicated method declarations. The main drawback is that we need to import the new `CreatedRulesetRef` trait.

What do you think @andreaphylum and @cd-work ?

### DIFF
--- a/examples/sandboxer.rs
+++ b/examples/sandboxer.rs
@@ -3,8 +3,8 @@
 
 use anyhow::{anyhow, bail};
 use landlock::{
-    Access, AccessFs, BitFlags, PathBeneath, PathFd, PathFdError, Ruleset, RulesetCreatedAttr,
-    RulesetError, RulesetStatus, ABI,
+    Access, AccessFs, BitFlags, PathBeneath, PathFd, PathFdError, Ruleset, RulesetAttr,
+    RulesetCreatedAttr, RulesetError, RulesetStatus, ABI,
 };
 use std::env;
 use std::ffi::OsStr;
--- a/examples/sandboxer.rs
+++ b/examples/sandboxer.rs
@@ -3,8 +3,8 @@
 
 use anyhow::{anyhow, bail};
 use landlock::{
-    Access, AccessFs, BitFlags, PathBeneath, PathFd, PathFdError, Ruleset, RulesetError,
-    RulesetStatus, ABI,
+    Access, AccessFs, BitFlags, PathBeneath, PathFd, PathFdError, Ruleset, RulesetCreatedAttr,
+    RulesetError, RulesetStatus, ABI,
 };
 use std::env;
 use std::ffi::OsStr;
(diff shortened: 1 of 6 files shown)


cd-work (approved):
I'm a fan. This seems pretty simple and the additional import is negligible considering the Rust compiler recommends it anyway even if it's missing.

l0kod:
I think I'll also implement `RulesetRef` to get a consistent builder pattern, with clearly defined transitions.

andreaphylum:
I like it better than #19! It looks simpler and more maintainable.

Only addition I'd recommend is sealing the trait, so it can't be implemented by other crates. It should be doable by renaming the current `RulesetCreatedRef` to `RulesetCreatedRefInner`, making it private, and creating a new `pub trait RulesetCreatedRef: RulesetCreatedRefInner {}`. Not super critical, but I think it'd be a good idea to avoid exposing unnecessary API surface.

l0kod:
I renamed the traits with *Attr (instead of *Ref), which makes more sense.

I'll do the same for the `Compatible` trait, but [after its refactoring](https://github.com/landlock-lsm/rust-landlock/pull/12/commits/43cc2c1ab5cbb9fde71f1f51998252eb62970831) because it will be easier to implement `AsMut<CompatLevel>`.

l0kod:
> Only addition I'd recommend is sealing the trait, so it can't be implemented by other crates. It should be doable by renaming the current `RulesetCreatedRef` to `RulesetCreatedRefInner`, making it private, and creating a new `pub trait RulesetCreatedRef: RulesetCreatedRefInner {}`. Not super critical, but I think it'd be a good idea to avoid exposing unnecessary API surface.

I't a good suggestion, but it is almost the case because `Ruleset` and `RulesetCreated` contain private fields, which makes them impossible to implement outside this crate. An external type could only implement the related `AsRef` and `AsMut` traits for these internal types, but in this case the *Attr traits would still make sense.

cd-work:
@l0kod Do you think it would make sense to upload a new version to crates.io? The last 0.1.0 version on it seems to be more of a placeholder and it rather useless compared to all the progress the git repository has made. With the current master pushed to crates.io, the published version would actually be usable.

Or do you have any major concerns with people using the current version of this crate?

l0kod:
> @l0kod Do you think it would make sense to upload a new version to crates.io? The last 0.1.0 version on it seems to be more of a placeholder and it rather useless compared to all the progress the git repository has made. With the current master pushed to crates.io, the published version would actually be usable.
> 
> Or do you have any major concerns with people using the current version of this crate?

The 0.1.0 on crates.io is indeed a placeholder. I want to merge #12 (with the new `set_compatibility()`) before creating a new version because this will also break the current API. I need to update the doc and rebase it but overall it is almost finished. Feel free to take a look. :wink:

PR #25, opened 2023-01-04
Use OwnedFd and AsFd instead of AsRawFd

bjorn3:
See https://rust-lang.github.io/rfcs/3128-io-safety.html for the motivation behind this change.

Signed-off-by: Björn Roy Baron <bjorn3_gh@protonmail.com>

Opened as separate PR as requested at https://github.com/landlock-lsm/rust-landlock/pull/22#issuecomment-1370947341

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,3 +8,5 @@ More information about Landlock can be found in the [official website](https://l
 
 This Rust crate provides a safe abstraction for the Landlock system calls along with some helpers.
 See the [Rust Landlock API documentation](https://landlock.io/rust-landlock/landlock/).
+
+Minimum Supported Rust Version (MSRV): 1.63
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ authors = [
 	"Vincent Dagonneau <vincentdagonneau@gmail.com>"
 ]
 edition = "2018"
+rust-version = "1.63"
 description = "Landlock LSM helpers"
 homepage = "https://landlock.io"
 repository = "https://github.com/landlock-lsm/rust-landlock"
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -8,3 +8,7 @@ More information about Landlock can be found in the [official website](https://l
 
 This Rust crate provides a safe abstraction for the Landlock system calls along with some helpers.
 See the [Rust Landlock API documentation](https://landlock.io/rust-landlock/landlock/).
+
+## Minimum Supported Rust Version (MSRV)
+
+Changing the MSRV is considered a breaking change and as such will result in a minor version bump. The current MSRV is 1.63.
```
bjorn3, line 14: I assumed you would consider changing the MSRV a breaking change.

l0kod: I found that we can specify this version (just after the `edition` entry) in Cargo.toml: https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-rust-version-field

With this, the readme changes are not necessary.

l0kod:
Please prefix this commit subject with "fs: "

README.md
bjorn3: I considered adding it, but given thay this key is only available from rust 1.56, it is effectively useless until we bump the MSRV to a later version.

bjorn3:
Done

README.md
l0kod: Indeed, we should keep the readme change, but the `rust-version` config is still useful for Rust >= 1.56, and it will display a warning because of unknown Cargo config key for old versions, which gives a hint too.

Please also only add this paragraph to the readme:
> Minimum Supported Rust Version (MSRV): 1.63

bjorn3: Right, 1.56 < 1.63 :facepalm:

bjorn3: Done

l0kod:
Thanks @bjorn3!

PR #29, opened 2023-01-06
Make AsFd compatible with Rust 1.63

l0kod:
The std::os::fd module is only available with Rust 1.66, but before that and starting with Rust 1.63, only the std::os::unix::io module is available.

Fixes #25

Cc @bjorn3


bjorn3:
Right, my bad. It is confusing that the docs for `std::os::fd` mention 1.63.0 as stabilization date.

PR #32, opened 2023-01-27
landlock: Bump to 0.2.0

l0kod:
Release a new crate version for current users before bringing more invasive API breaking changes (see #12 and #23).  The API is not ready to be stable yet but perfect is the enemy of good.

Switch to the 2021 Rust edition.

Remove the "authors" field.

Add four crate categories:
- api-bindings
- os::linux-apis
- virtualization
- filesystem

There is a limited set of categories and these ones make the more sense to me for now.

Cc @cd-work, @andreaphylum, @bjorn3, @rusty-snake, @r3dlight, @vdagonneau

Cf. phylum-dev/birdcage#19

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,10 +1,6 @@
 [package]
 name = "landlock"
-version = "0.1.0"
-authors = [
-	"Mickaël Salaün <mic@digikod.net>",
-	"Vincent Dagonneau <vincentdagonneau@gmail.com>"
-]
+version = "0.2.0"
 edition = "2021"
 rust-version = "1.63"
 description = "Landlock LSM helpers"
@@ -12,6 +8,7 @@ homepage = "https://landlock.io"
 repository = "https://github.com/landlock-lsm/rust-landlock"
 license = "MIT OR Apache-2.0"
 keywords = ["access-control", "linux", "sandbox", "security"]
+categories = ["api-bindings", "os::linux-apis", "virtualization", "filesystem"]
 exclude = [".gitignore"]
 readme = "README.md"
 


Cargo.toml
```diff
@@ -1,17 +1,15 @@
...
-version = "0.1.0"
-authors = [
-	"Mickaël Salaün <mic@digikod.net>",
-	"Vincent Dagonneau <vincentdagonneau@gmail.com>"
-]
-edition = "2018"
+version = "0.2.0"
+edition = "2021"
 rust-version = "1.63"
+authors = ["Mickaël Salaün <mic@digikod.net>"]
```
bjorn3, line 6: The authors field has been made optional: https://rust-lang.github.io/rfcs/3052-optional-authors-field.html

> The crates.io registry does not allow users to change the contents of already published versions: this is highly desirable to ensure working builds don't break in the future, but it also has the unfortunate side-effect of preventing people from updating the list of crate authors defined in Cargo.toml's package.authors field.
>
> This is especially problematic when people change their name or want to remove their name from the Internet, and the crates.io team doesn't have any way to address that at the moment except for deleting the affected crates or versions altogether. We don't do that lightly, but there were a few cases where we were forced to do so.
>
> The contents of the field also tend to scale poorly as the size of a project grows, with projects either making the field useless by just stating "The $PROJECT developers" or only naming the original authors without mentioning other major contributors.

l0kod: Good catch, I'll remove this field.

cd-work (approved):
Awesome, thanks.

I'm not that much in a hurry so I would have been fine waiting for #12, but it's probably not a terrible idea to make a release **before** a large API change either since that gives some more time for feedback to trickle in.

l0kod:
New release: https://crates.io/crates/landlock/0.2.0

l0kod:
https://github.com/landlock-lsm/rust-landlock/releases/tag/v0.2.0

Cc @micheleberetta98

PR #33, opened 2023-01-27
Switch to the 2021 Rust edition

l0kod:
Add changes required for the CI to pass.

Make sure any change is compatible with Rust 1.63, the current MSRV.

### DIFF
--- a/src/compat.rs
+++ b/src/compat.rs
@@ -116,11 +116,11 @@ lazy_static! {
             if ABI::is_known(n) || n == 0 {
                 ABI::from(n)
             } else {
-                panic!("Unknown ABI: {}", n);
+                panic!("Unknown ABI: {n}");
             }
         }
         Err(std::env::VarError::NotPresent) => ABI::iter().last().unwrap(),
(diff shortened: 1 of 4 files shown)


l0kod:
Cc #25 and #32

bjorn3 (approved):

src/compat.rs
```diff
@@ -116,11 +116,11 @@ lazy_static! {
             if ABI::is_known(n) || n == 0 {
                 ABI::from(n)
             } else {
-                panic!("Unknown ABI: {}", n);
+                panic!("Unknown ABI: {n}");
```
bjorn3, line 119: The error message for this should be fixed once you switch to the 2021 edition. In the 2015 and 2018 edition a panic with a single argument would use the given argument as a literal value to pass to `std::panic::panic_any`. With the 2021 edition it will use `format!()` for single argument `panic!()` too. (`std::panic::panic_any` was introduced to get the old behavior if necessary)

l0kod: Indeed, thanks!

PR #38, opened 2023-08-02
Initial duplicate RulesetCreated implementation

int5-grey:
This is a simple extension to the RulesetCreated object to allow for the object to be duplicated to a potential down-level consumer that can apply the ruleset in another thread or process without consuming the initial Ruleset. This is helpful when you have builders that will want to contain a parent ruleset copy that can be applied multiple times to new forks/threads without the need to reconstruct the builder.

 It doesn't override clone to do this as there is not a great pattern for handling errors if we take that approach and `fcntl` can fail for a variety of reasons.

Tested on `Ubuntu 23.04 6.2.0-20-generic`.

### DIFF
--- a/src/ruleset.rs
+++ b/src/ruleset.rs
@@ -653,6 +653,25 @@ impl RulesetCreated {
         };
         Ok(body()?)
     }
+
+    /// Creates a new `RulesetCreated` instance by duplicating the underlying file descriptor.
+    /// Rule modification will affect both `RulesetCreated` instances simultaneously.
+    ///
+    /// On error, returns [`std::io::Error`].
+    pub fn try_clone(&self) -> std::io::Result<Self> {
+        Ok(RulesetCreated {
+            fd: match self.fd {
+                -1 => -1,
+                self_fd => match unsafe { libc::fcntl(self_fd, libc::F_DUPFD_CLOEXEC, 0) } {
+                    dup_fd if dup_fd >= 0 => dup_fd,
+                    _ => return Err(Error::last_os_error()),
+                },
+            },
+            no_new_privs: self.no_new_privs,
+            requested_handled_fs: self.requested_handled_fs,
+            compat: self.compat,
+        })
+    }
 }
 
 impl Drop for RulesetCreated {
--- a/src/compat.rs
+++ b/src/compat.rs
@@ -255,7 +255,7 @@ fn compat_state_update_2() {
 }
 
 #[cfg_attr(test, derive(Debug, PartialEq))]
-#[derive(Clone)]
+#[derive(Copy, Clone)]
 pub(crate) struct Compatibility {
     abi: ABI,
     pub(crate) level: Option<CompatLevel>,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -343,4 +343,21 @@ mod tests {
             false,
         );
     }
+
+    #[test]
+    fn ruleset_created_try_clone() {
+        check_ruleset_support(
+            ABI::V1,
+            Some(ABI::V1),
+            move |ruleset: Ruleset| -> _ {
+                Ok(ruleset
+                    .handle_access(AccessFs::Execute)?
+                    .create()?
+                    .add_rule(PathBeneath::new(PathFd::new("/")?, AccessFs::Execute))?
+                    .try_clone()?
+                    .restrict_self()?)
+            },
+            false,
+        );
+    }
 }


src/ruleset.rs
```diff
@@ -491,14 +491,32 @@ impl RulesetCreated {
         }
     }
 
+    /// Creates a copy of this object and duplicates the file descriptor with
+    /// CLOEXEC set.
+    ///
+    /// On error, returns a wrapped [`RulesetError::Duplicate`].
+    pub fn duplicate(&self) -> Result<RulesetCreated, RulesetError> {
```
bjorn3, line 498: `std::io::File` calls this `try_clone`.

l0kod: The issue with the `clone` name is that developers would think it is a standalone object, which is not the case because the kernel object, referenced by the ruleset FD, would be the same. This means that the one of these object can affect the other.

This `duplicate` method should probably be renamed to something like `new_ref()`, and be part of the `RulesetCreatedAttr` trait. I'm wondering if it wouldn't be safer to keep a reference to the original object to avoid unexpected changes. :thinking:  Would that be an issue for your use case?

src/ruleset.rs
```diff
@@ -491,14 +491,32 @@ impl RulesetCreated {
...
+
     /// Attempts to restrict the calling thread with the ruleset
     /// according to the best-effort configuration
     /// (see [`RulesetCreated::set_best_effort()`]).
     /// Call `prctl(2)` with the `PR_SET_NO_NEW_PRIVS`
     /// according to the ruleset configuration.
     ///
     /// On error, returns a wrapped [`RestrictSelfError`].
-    pub fn restrict_self(mut self) -> Result<RestrictionStatus, RulesetError> {
+    pub fn restrict_self(&mut self) -> Result<RestrictionStatus, RulesetError> {
```
l0kod, line 519: I need to ponder this change a bit. This designed was initially chosen to avoid unexpected behavior, but this use case could change it.

l0kod (commented):
Thanks Carl!

We should not create another `RulesetCreated` object but something like `RulesetRef` that would not allow modification of the original `RulesetCreated` object (i.e. no `RulesetCreatedAttr` implementation). The `restrict_self` method should be part of a new trait, implemented by `RulesetCreated` and `RulesetRef`.

There is currently no way to remove the write mode of the ruleset FD, but that would make sense to be able to drop this permission. I'll work on that on the kernel side, but we should probably not wait for this new kernel feature to land.

src/ruleset.rs
int5-grey: No it wouldn't, I'll work on those changes.

src/ruleset.rs
```diff
@@ -558,20 +572,69 @@ impl RulesetCreated {
     }
 }
 
-impl Drop for RulesetCreated {
-    fn drop(&mut self) {
-        if self.fd >= 0 {
-            unsafe { close(self.fd) };
+macro_rules! define_rulset_restrict_object {
```
l0kod, line 575: Can this macro be replaced with a trait?

src/ruleset.rs
```diff
@@ -558,20 +572,69 @@ impl RulesetCreated {
...
-impl AsMut<RulesetCreated> for RulesetCreated {
-    fn as_mut(&mut self) -> &mut RulesetCreated {
-        self
-    }
+        impl $t {
+            #[allow(unused)]
+            fn new(ruleset: Ruleset, fd: RawFd) -> Self {
+                $t {
+                    fd,
+                    no_new_privs: true,
```
l0kod, line 591: I'll merge #12 in a few day and this will change some code (e.g. no_new_privs), but I want to update the CI tests before (cf. #41). You can already rebase on #12.

l0kod (commented):
Can you please squash the commits that change the same code, especially those renaming variables in new code (e.g. Duplicate/NewRef)? This makes the review easier and also ease bisecting.

I'd like to keep the current commit message convention:
* short subject with a prefix
* 72-column formatted body

You need to add your Signed-off-by (see the failed DCO check).

All commits must pass the CI tests.

l0kod:
Can you please rebase on the main branch, squash into one commit, and fix the CI issues?

src/ruleset.rs
```diff
@@ -551,6 +553,34 @@ pub trait RulesetCreatedAttr: Sized + AsMut<RulesetCreated> + Compatible {
...
+    /// On error, returns a wrapped [`RulesetError::NewRef`].
+    fn new_ref(&self) -> Result<RulesetCreated, RulesetError> {
+        let self_ref = self.as_ref();
+
+        let dup_fd = unsafe { libc::fcntl(self_ref.fd, libc::F_DUPFD_CLOEXEC) };
+        if dup_fd == -1 {
+            return Err(RulesetError::NewRef(Error::last_os_error()));
+        }
+
+        Ok(RulesetCreated {
```
l0kod, line 577: You can use `RulesetCreated::new()` instead, wrap this new ruleset into a `RulesetRef` object, and implement a `restrict_self()` method for `RulesetRef`. This way, no new trait and fully compatible with existing code. The documentation of this new `restrict_self()` method should just point to the `RulesetCreated::restrict_self()` one.

src/ruleset.rs
```diff
@@ -551,6 +553,34 @@ pub trait RulesetCreatedAttr: Sized + AsMut<RulesetCreated> + Compatible {
         <Self as AsMut<RulesetCreated>>::as_mut(&mut self).no_new_privs = no_new_privs;
         self
     }
+
+    /// Creates a copy of this object and duplicates the file descriptor with
+    /// CLOEXEC set. The `RulesetRef` object that is produced is intended
+    /// to be a readonly copy of the `RulesetCreated` object, however this
+    /// is not currently supported by the landlock kernel module. The intention
```
l0kod, line 560: We should add a warning to make sure nodoby use this new method for security purpose because it would be easy to access the underlying file descriptor and update the related ruleset.

l0kod (changes requested):
Please fix issues reported by the CI.

src/ruleset.rs
```diff
@@ -410,7 +410,9 @@ impl Compatible for RulesetCreated {}
 
 impl Compatible for &mut RulesetCreated {}
 
-pub trait RulesetCreatedAttr: Sized + AsMut<RulesetCreated> + Compatible {
+pub trait RulesetCreatedAttr:
+    Sized + AsMut<RulesetCreated> + AsRef<RulesetCreated> + Compatible
+{
```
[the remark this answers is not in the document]
int5-grey: `new_ref` is part of the trait, which needs to access the file descriptor declared in RulesetCreated.

src/ruleset.rs
```diff
@@ -669,6 +699,12 @@ impl AsMut<RulesetCreated> for RulesetCreated {
     }
 }
 
+impl AsRef<RulesetCreated> for RulesetCreated {
```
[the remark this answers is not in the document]
int5-grey: See above, its to access the file descriptor in RulesetCreated

src/ruleset.rs
int5-grey: RulesetRef is gone now in the latest, I think it makes sense to just move new_ref to the RulesetCreated object itself, no need for a trait. I will also remove the AsRef from the attr trait.

src/ruleset.rs
```diff
@@ -719,6 +745,29 @@ fn ruleset_created_attr() {
     );
 }
 
+#[test]
+fn ruleset_created_new_ref() {
+    Ruleset::from(ABI::V1)
```
l0kod, line 750: Tests creating real rulesets needs to use the `check_ruleset_support()` helper and be in `lib.rs`. This is why running the test with Linux 5.10 failed (because Landlock is not supported with this version).

l0kod (changes requested):
As reported by the CI, you need to add a Signed-off-by tag in the commit message.

src/ruleset.rs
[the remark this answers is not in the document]
l0kod: Is it still needed with your last version?

src/ruleset.rs
```diff
@@ -653,6 +653,32 @@ impl RulesetCreated {
...
+    /// behind this method is to provide a way to keep a parent ruleset intact so
+    /// it can be applied to any number of new processes or threads without
+    /// creating the object over and over.
+    ///
+    /// Any modification of the underlying ruleset changes will alter both
+    /// the new RulesetCreated and the original.
+    ///
+    /// On error, returns a wrapped [`RulesetError::NewRef`].
+    pub fn new_ref(&self) -> Result<RulesetCreated, std::io::Error> {
+        let dup_fd = unsafe { libc::fcntl(self.fd, libc::F_DUPFD_CLOEXEC, 0) };
```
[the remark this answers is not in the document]
int5-grey: I changed to `try_clone`. I think changing to `OwnedFd` is a good idea, but it will require more work to make the rest of the compat state work correctly since `OwndFd` can't be an invalid descriptor which is what happens now with a few states. This is out of scope of this particular work item (for me at least).

src/ruleset.rs
```diff
@@ -653,6 +653,36 @@ impl RulesetCreated {
...
         Ok(body()?)
     }
+
+    /// Creates a copy of this object and duplicates the file descriptor with
+    /// CLOEXEC set. The `RulesetCreated` object that is produced is intended
+    /// to be a readonly copy of the original object, however this
+    /// is not currently supported by the landlock kernel module. The intention
+    /// behind this method is to provide a way to keep a parent ruleset intact so
+    /// it can be applied to any number of new processes or threads without
+    /// creating the object over and over.
```
l0kod, line 663: This documentation is not accurate anymore.

src/ruleset.rs
```diff
@@ -653,6 +653,36 @@ impl RulesetCreated {
...
+    ///
+    /// Any modification of the underlying ruleset changes will alter both
+    /// the new RulesetCreated and the original.
+    ///
+    /// On error, returns [`std::io::Error`].
+    pub fn try_clone(&self) -> Result<Self, std::io::Error> {
```
l0kod, line 669: Can you please return `Result<std::io::Error>` instead and update the test accordingly?

l0kod (changes requested):

src/ruleset.rs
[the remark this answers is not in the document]
l0kod: OK, I'll work on this `OwnedFd` change after your PR.

l0kod:
[The DCO check failed.](https://github.com/landlock-lsm/rust-landlock/pull/38/checks?check_run_id=22943335268)

src/ruleset.rs
int5-grey: Do you mean return a `std::io::Result<Self>`?

l0kod (approved):

l0kod:
I did some cosmetic changes but this looks good. Thanks!

PR #45, opened 2023-08-30
 Make sure the same Cargo version is used for all tests

l0kod:
Set and export the CARGO environment variable to make sure the ubuntu_22_rust_stable job uses a consistent Cargo version for landlock-test-tools/test-rust.sh tests and the native test.


l0kod:
This completes #41

PR #47, opened 2023-08-31
Run tests built with Rust MSRV

l0kod:
This is useful to detect test inconsistencies between Rust versions.

This new MSRV/UML test would have failed without commit dbd4105e0bed ("src: Execute each restrict_self() test in a dedicated thread"): #46 

Use Ubuntu 22.04 to run MSRV and UML tests.

Sort steps to move all the checkout actions early.


l0kod:
Dependency: #41

PR #55, opened 2024-01-10
Add initial networking support

l0kod:
Linux 6.7 adds an initial Landlock network support with TCP bind and connect restrictions.


l0kod:
I still need to address a few TODOs but the main changes should work fine.

I'd like to leverage the type system for the port check (instead of the runtime check) but I'm still thinking about a proper API.

l0kod:
> I'd like to leverage the type system for the port check (instead of the runtime check) but I'm still thinking about a proper API.

I switched to `u16` ports an removed the `PortOverflow` error. We'll add new methods with other integer types (and related overflow error) if needed.

l0kod:
I'll release a new version of this crate tomorrow if no issue is reported.

l0kod:
That was a long ride, but it's now [published](https://github.com/landlock-lsm/rust-landlock/releases/tag/v0.4.0)! :partying_face:

As you can see, this was due to some refactoring to generalize the internal library code for the new class of access rights and rule (network). This investment will greatly simplify the implementation of future changes.

As described in the [changelog](https://github.com/landlock-lsm/rust-landlock/blob/main/CHANGELOG.md#v040), you'll need to update `Cargo.toml` with the `0.4.0` version of this crate.

PR #69, opened 2024-06-19
Bump to 0.4.0

l0kod:
This is a breaking change because of #66, and this changes a bit the current behavior because of #67.


l0kod:
See https://github.com/landlock-lsm/rust-landlock/pull/55#issuecomment-2178206380

PR #70, opened 2024-07-28
Fix typo in doc-comment

rusty-snake:


l0kod:
Thanks for the fix. Could you please add a `Signed-off-by` tag and prefx the commit's subject with "net: "?

PR #76, opened 2024-10-22
Add CONTRIBUTING.md doc

dburgener:
This is essentially to make the comment over current_kernel_abi() more visible and add some handholding for new developers.  Coming in and running this crate on an older kernel results in multiple test failures, and finding the current_kernel_abi() comment takes a little investigation, so spelling this all out for new contributors may help save some time.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,21 @@
+# Contributing
+
+Thanks for your interest in contributing to rust-landlock!
+
+## Testing vs kernel ABI
+
+The Landlock functionality exposed differs between kernel versions.  In order
+to test all possible variations, the rust-landlock tests will run different
+subsets of tests based on the landlock support in the current kernel.
+
+In order to fully test a change, it should be verified against a range of
+kernel versions.  This is done by the github actions CI, but is currently
+challenging to do locally.  For local development, running `cargo test` will
+test against your currently running kernel version, which may result in some
+tests being skipped.
+
+The kernel to test against can be overridden using the LANDLOCK_CRATE_TEST_ABI
+environmental variable.  For more details see the comment in
+`compat.rs:current_kernel_abi()`.
+
+For more information about Landlock ABIs see https://landlock.io/rust-landlock/landlock/enum.ABI.html
--- a/src/compat.rs
+++ b/src/compat.rs
@@ -146,7 +146,7 @@ lazy_static! {
                 panic!("Unknown ABI: {n}");
             }
         }
-        Err(std::env::VarError::NotPresent) => ABI::iter().last().unwrap(),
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,28 @@
...
+You should use the ABI matching your actual kernel version.  The test
+`current_kernel_abi()` verifies that the ABI you set in the variable matches
+your kernel version.
+
+If LANDLOCK_CRATE_TEST_ABI is unset, it defaults to the latest ABI supported by
+rust-landlock.
+
+Note that if you are running with older kernels, you will be missing some
+tests, which could cause a difference between your local testing and the
+github actions CI.  The github actions CI tests against all supported kernel
```
dburgener, line 25: I think the claim that github checks all supported ABIs is true, but the github actions here is fairly complicated, so I struggled to fully verify it.  Can you double check me on that?

l0kod: This is correct, GitHub actions test each commit against each meaningful kernel versions.

l0kod (commented):
Thanks for contributing! :upside_down_face: 

The main issue with this new documentation is that for a change to be fully tested, each commit needs to be tested against all meaningful kernel version, hence the complex GitHub action. We should probably improve [landlock-test-tools](https://github.com/landlock-lsm/landlock-test-tools?tab=readme-ov-file#rust-landlock) to easily loop through all the available kernel and run all the tests. For now, the initial kernel are embedded in landlock-test-tools, which is not great, and this was replaced for GitHub action with a more dynamic approach: #72. However, it would require more work to be able to easily run the same tests locally (e.g. building and caching several kernel versions), and it would significantly increase test duration.

By default, if `LANDLOCK_CRATE_TEST_ABI` is not set, tests assume we are testing the latest kernel version. We should probably change this default behavior to run tests against the version of the running kernel, which will simplify this doc too.

dburgener:
I pushed an update that tries to address your comment above.  Let me know what you think.

l0kod:
Sorry for the late reply. This looks good. Could you please squash the two commits about contributing and move them to the top. Also, the current convention is to have commit subject with a prefix (I should add that to the doc BTW), something like:
* compat: Set default LANDLOCK_CRATE_TEST_ABI to the current kernel
* contributing: Add initial documentation

dburgener:
> and move them to the top. 

Can you clarify what you mean about "the top"?  Different views of git logs order commits differently.  Would you like the doc change to apply in git history before or after the compat change?

l0kod:
> > and move them to the top.
> 
> Can you clarify what you mean about "the top"? Different views of git logs order commits differently. Would you like the doc change to apply in git history before or after the compat change?

Just to have the `LANDLOCK_CRATE_TEST_ABI` change before the new documentation.

PR #78, opened 2024-12-13
Update to thiserror version 2.

dburgener:
The breaking changes between version 1 and 2 do not affect rust-landlock's usage of thiserror.


l0kod:
Thanks!

Issue #57, opened 2024-02-21
Add system tests


l0kod:
We currently only rely on the kernel to enforce restrictions, and we only have unit tests. We need system tests to make sure sandboxing is restricting as expected.

Things to consider: #41 and #46

n0toose:
Hi, would it make sense to create a new crate with test programs, and have a "wrapper" in the test suite build those test programs and run them?

My plan would be to follow a similar approach to that used in https://github.com/hermit-os/uhyve/blob/main/tests/common.rs, only that we wouldn't build and execute *unikernels* here. There isn't a stabilized cargo API to use yet (although that was offered as a GSoC project this year).

PR #92, opened 2025-04-16
tests: add case for AccessFs::from_file

n0toose:
It's a minor change, but I've been having some problems with using `from_file` in Ubuntu 22.04 and wrote this to be extra sure that I wasn't overseeing something super trivial. Thought it might turn out to be useful later down the line.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -156,8 +156,10 @@ fn consistent_access_fs_rw() {
         let access_all = AccessFs::from_all(abi);
         let access_read = AccessFs::from_read(abi);
         let access_write = AccessFs::from_write(abi);
+        let access_file = AccessFs::from_file(abi);
         assert_eq!(access_read, !access_write & access_all);
         assert_eq!(access_read | access_write, access_all);
+        assert_eq!(access_file, access_all & ACCESS_FILE);
     }
 }
 


src/fs.rs
```diff
@@ -190,7 +192,6 @@ impl PrivateAccess for AccessFs {
 }
 
 // TODO: Make ACCESS_FILE a property of AccessFs.
-// TODO: Add tests for ACCESS_FILE.
```
l0kod, line 193: This TODO is a reminder to check that `ACCESS_FILE` can be used on a file *and* that is not the case for its complement. For instance, if `IoctlDev` wasn't part of `ACCESS_FILE`, such a test should detect it. That would not be a unit test but an integration test, see #57.

src/fs.rs
```diff
@@ -154,8 +154,10 @@ fn consistent_access_fs_rw() {
         let access_all = AccessFs::from_all(abi);
         let access_read = AccessFs::from_read(abi);
         let access_write = AccessFs::from_write(abi);
+        let access_file = AccessFs::from_file(abi);
         assert_eq!(access_read, !access_write & access_all);
         assert_eq!(access_read | access_write, access_all);
+        assert_eq!(access_file, access_all & ACCESS_FILE);
```
l0kod, line 160: This is exactly what does `from_file()` but we can add that to tests.

n0toose: Course of action unclear. Followed a similar approach to `assert_eq!(access_read, !access_write & access_all);` ("regression test").

l0kod:
Thanks!

Issue #77, opened 2024-11-02
Improve "file names that don't exist" case explanation


n0toose:
The documentation mentions a case of ["file names that don’t exist"](https://docs.rs/landlock/latest/landlock/#compatibility). This got me thinking. Is it possible to whitelist a specific "path", despite no file (and therefore, no file descriptor) for a regular path to be created, without giving R/W permissions to the parent folder? (For example, so as to create a `log.txt` file in the current working directory, without the process being able to inspect anything else in the current working directory apart from the log?)

https://github.com/landlock-lsm/rust-landlock/blob/94721d26b2fd1151e71bd7a3aa5a43c463a22347/src/uapi/landlock.rs#L15

https://github.com/landlock-lsm/rust-landlock/blob/94721d26b2fd1151e71bd7a3aa5a43c463a22347/src/fs.rs#L74
-
https://github.com/landlock-lsm/rust-landlock/blob/94721d26b2fd1151e71bd7a3aa5a43c463a22347/src/fs.rs#L129

Derived from the examples:

```
pub fn initialize_landlock() -> Result<RestrictionStatus, LandlockRestrictError> {
	let abi = ABI::V5;
	let access_all: landlock::BitFlags<AccessFs, u64> = AccessFs::from_all(abi);
	let access_read: landlock::BitFlags<AccessFs, u64> = AccessFs::from_read(abi);

	Ok(Ruleset::default()
		.handle_access(access_all)?
		.create()?
		.add_rules(
			WHITELISTED_PATHS
				.get()
				.unwrap()
				.as_slice()
				.iter()
				.map::<Result<_, LandlockRestrictError>, _>(|p| {
					Ok(PathBeneath::new(PathFd::new(p)?, access_all))
				}),
		)?
		.restrict_self()?)
}
```

l0kod:
The paths are all opened and the kernel actually only look for the related inodes. It is then not possible to create rules based on inexistant file.

We should probably change the documentation with "files that don't exist".

PR #94, opened 2025-04-17
docs: add more background to PathBeneath example

n0toose:
This example is shown at the homepage of rust-landlock's documentation, and essentially serves as some sort of an introduction to implementers. Since this example serves as some sort of a "first impression", it is quite easy for the reader to make a false generalization.

To fix this, some additional background regarding the compatibility of rules was provided. Specifically, CompatLevel. I did not go through the effort of mentioning CompatLevel in its entirety, but I explained enough to show that the behavior of Landlock ignoring incompatible rules is not inherent to Landlock, and provided a counter-example with a link for the reader.

Implicitly, this somewhato resolves the ambiguity of the comment:

```
// The FullyEnforced case must be tested by the developer.
```

---

Partially fixes https://github.com/landlock-lsm/rust-landlock/issues/77.

### DIFF
--- a/src/fs.rs
+++ b/src/fs.rs
@@ -552,9 +552,16 @@ fn path_fd() {
 ///
 /// # Note
 ///
-/// Landlock rules operate on file descriptors, not paths. This is a helper to create rules based on paths. **It
-/// silently ignores paths that cannot be opened**, and automatically adjusts access rights according to file types when
-/// possible.
+/// From the kernel's perspective, Landlock rules operate on file descriptors, not paths.
+/// This is a helper to create rules based on paths. Here, `path_beneath_rules()` silently ignores
+/// paths that cannot be opened, hence making the obtainment of a file descriptor impossible. When
+/// possible and for a given path, `path_beneath_rules()` automatically adjusts [access rights](`AccessFs`),
+/// depending on whether a directory or a file is present at that said path.
+///
+/// This behavior is the result of [`CompatLevel::BestEffort`], which is the default compatibility level of
+/// all created rulesets. Thus, it applies to the example below. However, if [`CompatLevel::HardRequirement`]
+/// is set using [`Compatible::set_compatibility`], attempting to create an incompatible rule at runtime will cause
+/// this crate to raise an error instead.
 ///
 /// # Example
 ///


src/fs.rs
```diff
@@ -546,9 +546,16 @@ fn path_fd() {
 ///
 /// # Note
 ///
-/// Landlock rules operate on file descriptors, not paths. This is a helper to create rules based on paths. **It
-/// silently ignores paths that cannot be opened**, and automatically adjusts access rights according to file types when
-/// possible.
+/// This is a helper to create rules based on paths. However, Landlock rules operate on file descriptors, not paths.
+/// When defining a rule based on a path, a file should be present at the provided path. Otherwise, Landlock
+/// **silently ignores paths that cannot be opened by default**, as it deems such rules to be incompatible.
+/// When possible, Landlock automatically adjusts access rights of paths that can be opened,
```
n0toose, line 552: Irrelevant to this PR, but, as far as I am aware, the automatic adjustment of access rights *only* applies to `CompatLevel::BestEffort` and not `CompatLevel::HardRequirement`. Perhaps this could be documented more prominently.

n0toose: If my assumption is not wrong, this actually has implications for when using `from_read` and `from_write` against files - thus requiring the user to AND with `AccessFs::from_file` (or `ACCESS_FILE`) - if I am not wrong, and a new issue should be opened. I also made a loosely related PR: https://github.com/landlock-lsm/rust-landlock/pull/92

l0kod: `from_read()` and `from_write()` only returns access rights that are in one category or the other. Developers can to a binary AND with `from_file()` to get the subset of access rights that make sense for a file. However, when using the default `BestEffort` mode, the `add_rule()` calls automatically adjust access rights according to the file type. If the developer explicitly configures the ruleset with `HardRequirement`, then an error is returned. See https://github.com/landlock-lsm/rust-landlock/blob/1f55adc26239435a347cc6766b7be42d30a1fd13/src/fs.rs#L287-L305

`HardRequirement` should not be use lightly, but only for tests and fully owned systems. `BestEffort` should be used by most sandboxed programs.

`PathBeneath::new(PathFd::new(path)?, access)` can be use to error out if a path doesn't exist.

src/fs.rs
```diff
@@ -546,9 +546,16 @@ fn path_fd() {
 ///
 /// # Note
 ///
-/// Landlock rules operate on file descriptors, not paths. This is a helper to create rules based on paths. **It
-/// silently ignores paths that cannot be opened**, and automatically adjusts access rights according to file types when
-/// possible.
+/// This is a helper to create rules based on paths. However, Landlock rules operate on file descriptors, not paths.
+/// When defining a rule based on a path, a file should be present at the provided path. Otherwise, Landlock
+/// **silently ignores paths that cannot be opened by default**, as it deems such rules to be incompatible.
```
l0kod, line 551: > Landlock **silently ignores paths that cannot be opened by default**

This is false. Landlock (the kernel feature) cannot ignore paths because it only takes file descriptors. Only the `path_beneath_rules()` helper ignores non-existing paths.

src/fs.rs
```diff
@@ -546,9 +546,16 @@ fn path_fd() {
...
+/// When possible, Landlock automatically adjusts access rights of paths that can be opened,
+/// based on the paths' files' file type.
+/// 
+/// This behavior is the result of [`CompatLevel::BestEffort`], which is the default compatibility level of
```
l0kod, line 555: This part is good.

l0kod (changes requested):

n0toose:
Hi, thanks for the insightful reviews.

> HardRequirement should not be use lightly, but only for tests and fully owned systems. BestEffort should be used by most sandboxed programs.

A thing that I have not really 100% understood throughout my discussions with you about HardRequirement is that, while I do get what it is intended for testing - not the preferred default for developers - I am not sure how to properly document that. Should I add a sidenote mentioning that HardRequirement is for tests, or just leave it at https://github.com/landlock-lsm/rust-landlock/pull/94#discussion_r2059987819 (which implicitly prompts the reader to dig deeper - but doesn't quite acknowledge the existence of HardRequirement)?

n0toose:
I tried to improve the state of the PR using the feedback you provided; I tried to be a bit more explicit in my writing (and imitate your style of writing - kind of) while trying not to write too many essays at once. I expect an iteration or two, but hope this is an improvement.

l0kod:
I updated it with small cosmetic changes/fixes. Thanks!

PR #102, opened 2025-04-24
Implement Deref instead of AsRef for Errno

l0kod:
It makes more sense and it makes it easier to dereference the inner value.

Import c_int to lighten code.

Improves #86


l0kod:
Requires to update https://github.com/landlock-lsm/landlockconfig/pull/18

Issue #110, opened 2025-10-02
Test failures on i386


alexanderkjall:
Hi, it looks like there is some test failures on i386 that look kind of important:

The full log is available here: https://ci.debian.net/data/autopkgtest/testing/i386/r/rust-landlock/64842386/log.gz

The important part is:

```
124s 
124s failures:
124s 
124s ---- uapi::landlock::bindgen_test_layout___kernel_fd_set stdout ----
124s 
124s thread 'uapi::landlock::bindgen_test_layout___kernel_fd_set' panicked at src/uapi/landlock.rs:44:5:
124s assertion `left == right` failed: Size of __kernel_fd_set
124s   left: 64
124s  right: 128
124s stack backtrace:
124s    0: __rustc::rust_begin_unwind
124s              at /usr/src/rustc-1.87.0/library/std/src/panicking.rs:697:5
124s    1: core::panicking::panic_fmt
124s              at /usr/src/rustc-1.87.0/library/core/src/panicking.rs:75:14
124s    2: core::panicking::assert_failed_inner
124s    3: core::panicking::assert_failed
124s              at /usr/src/rustc-1.87.0/library/core/src/panicking.rs:380:5
124s    4: landlock::uapi::landlock::bindgen_test_layout___kernel_fd_set
124s              at ./src/uapi/landlock.rs:44:5
124s    5: landlock::uapi::landlock::bindgen_test_layout___kernel_fd_set::{{closure}}
124s              at ./src/uapi/landlock.rs:41:41
124s    6: core::ops::function::FnOnce::call_once
124s              at /usr/src/rustc-1.87.0/library/core/src/ops/function.rs:250:5
124s    7: core::ops::function::FnOnce::call_once
124s              at /usr/src/rustc-1.87.0/library/core/src/ops/function.rs:250:5
124s note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
124s 
124s ---- uapi::landlock::bindgen_test_layout_landlock_net_port_attr stdout ----
124s 
124s thread 'uapi::landlock::bindgen_test_layout_landlock_net_port_attr' panicked at src/uapi/landlock.rs:231:5:
124s assertion `left == right` failed: Alignment of landlock_net_port_attr
124s   left: 4
124s  right: 8
124s stack backtrace:
124s    0: __rustc::rust_begin_unwind
124s              at /usr/src/rustc-1.87.0/library/std/src/panicking.rs:697:5
[...]

PR #111, opened 2025-10-02
Generate 64-bit and 32-bit bindings

l0kod:
See #110


bjorn3:
Wouldn't this break all non x86/x86_64 archs?

PR #120, opened 2026-04-10
Add ABI v7: log flags

l0kod:
Depends on #119

PR #122, opened 2026-05-19
Rename set_no_new_privs() to no_new_privs()

l0kod:
The Rust builder pattern is meant to read as a declarative list of configurations rather than a sequence of imperative commands, see https://rust-lang.github.io/api-guidelines/type-safety.html?#builders-enable-construction-of-complex-values-c-builder https://rust-lang.github.io/rfcs/0344-conventions-galore.html#gettersetter-apis

Dropping `set_` from boolean field setters reduces boilerplate and makes the chain read uniformly (with upcomming log_* setters):

    Ruleset::default()
        .handle_access(...)?
        .create()?
        .log_same_exec(false)?
        .log_new_exec(true)?
        .log_subdomains(false)?
        .no_new_privs(false)
        .restrict_self()?;

Note that this method should only be rarely called because of the default value.

Verb-led method names are reserved for distinct actions beyond field assignment (handle_access, add_rule, restrict_self).

Keep set_no_new_privs as a #[deprecated] thin wrapper calling no_new_privs() to preserve backward compatibility.

set_compatibility() is out of scope for for now, see https://github.com/landlock-lsm/rust-landlock/pull/58

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -115,6 +115,7 @@ mod errata;
 mod errors;
 mod fs;
 mod net;
+mod prctl;
 mod ruleset;
 mod scope;
 mod uapi;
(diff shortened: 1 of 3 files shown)


l0kod:
Related to #120 for consistency naming.

PR #130, opened 2026-06-23
Add ABI v8: TSYNC

l0kod:
Depends on #129


l0kod:
@jarkkojs, it would be useful if you could test it. I still need to do some polishing, but the main code should be OK.

jarkkojs:
Sure, can you give me day or two so I could do experimental change to Landstrip and get that way a real end-to-end test done?

l0kod:
> Sure, can you give me day or two so I could do experimental change to Landstrip and get that way a real end-to-end test done?

Sure, there is no rush.

jarkkojs:
> > Sure, can you give me day or two so I could do experimental change to Landstrip and get that way a real end-to-end test done?
> 
> Sure, there is no rush.

Here's what I plan to do.

First off, my test suite is not "real" cargo test:    

https://github.com/landstrip/landstrip/tree/main/tests

That is actually by design as it making it "deployable".

Secondly, I have somewhat robust kernel testing harness to which I can package Landstrip and test program: https://git.kernel.org/pub/scm/linux/kernel/git/jarkko/linux-tpmdd-test.git/

So I'll do that compile in modified Landstrip and can run quite complex set of tests, which should exercise pretty well at least part of the new features and I can add some more to [data.txt](https://github.com/landstrip/landstrip/blob/main/tests/data.txt) if I need extra coverage :-)

Thus there is a bit of preparation but nice thing is that it will end up to my regular Linux kernel QA cycle (because why I would remove it after all the extra trouble).

jarkkojs:
Tested-by: Jarkko Sakkinen <jarkko.sakkinen@iki.fi>

With non-production quality changes but it does the job. I'll do a proper integration later.

eslerm:
Responding to @l0kod's https://github.com/landlock-lsm/rust-landlock/pull/131#issuecomment-5037800849

Ran this on two bare-metal boxes, ABI v8 (Linux 7.0.13) and ABI v9 (Linux 7.1.3), both x86_64, from `482c107` with a clean tree. The suite is green on both, including the abi-4 (6.7) and abi-6 (6.12) rungs CI currently shows red.

Those two red rungs look like the UML flakiness the `abi_v8_all_threads` comment already calls out: both fail with `userspace - ptrace set fp regs failed` and a guest `Kernel panic - Attempted to kill init`, and the same rungs pass on the base commit `e64b5c5` (examples-only delta), so it reads as nondeterministic rather than a regression.

I ran a few additional checks outside of the suite, all clean on both, including:

- **no_new_privs propagation.** The sibling goes 0 -> 1 with `all_threads(true)` and stays 0 with `all_threads(false)`, so it's attributable to TSYNC and not to `restrict_self()` generally.
- **Spawn race.** Concurrent workers plus tens of thousands of threads spawned across the `restrict_self()` call, ~900k access attempts per run: no thread saw a successful access once enforcement had been observed, over 6 runs on each box. As a control, the same probe with one thread restricting only itself reports ~500-720k such escapes, so the check does fire when confinement is partial.

One small suggestion, on the error surface. On the domain-less builder, `all_threads(true)` without `log_subdomains(false)` surfaces as `failed to restrict the calling thread: Bad file descriptor (os error 9)`. I think not pre-checking is right, but the `ruleset_fd == -1` pairing lives only in the `all_threads()` rustdoc, which isn't where you look when debugging an errno. Since it's public API once released, naming the pairing at the error site would make the EBADF self-diagnosing. I hit it before I found the doc.

I can test PRs on 7.0.13 / 7.1.3 boxes whenever it's useful. I didn't cover i686 or the pre-v8 drop paths.

PR #131, opened 2026-07-11
bump to ABI v9, add TSYNC and FS_RESOLVE_UNIX

c-cube:
Hi! I've been using this crate for a little while and really appreciate it. I realized yesterday that connecting to unix sockets (including `tmux`'s control socket) is not restricted by landlock by default unless `LANDLOCK_ACCESS_FS_RESOLVE_UNIX` is used.

So, I'm mostly interested in this FS_RESOLVE_UNIX flag, but since it's in ABI v9 I'm also adding ABI v8 as I go.

I did not run tests locally yet, my kernel is 7.0.13. 

Disclaimer: this has been written with LLM assistance, but I reviewed every line.


eslerm:
Thanks for adding V8/V9. One heads-up on `e7eed33`: `cargo test` fails on `consistent_access_fs_rw`. `ResolveUnix` is added to both the read and write sets:

- `fs.rs:125` — `from_read(V9) = from_read(V8) | ResolveUnix`
- `fs.rs:150` — `from_write(V9) = from_write(V8) | ResolveUnix`

That test wants read and write disjoint (`access_read == !access_write & access_all`), so the shared bit trips it:

```
test fs::consistent_access_fs_rw ... FAILED
  left:  Execute | ReadFile | ReadDir | ResolveUnix
  right: Execute | ReadFile | ReadDir
```

It's const bitflag logic over `ABI::iter()`, so it fails on a plain `cargo test` anywhere, no special kernel. Nothing consumer-facing is wrong: `from_all(V9)` contains `ResolveUnix` either way and `resolve_unix_access` passes. Fix is just to keep the bit in one of the two sets. The kernel treats it as neither read nor write so which side is your call; `from_all` comes out identical. I kept it write-side in a local build and the unit suite goes green (76/76).

Since I'm pinning this to block pathname-socket connect, I ran the enforcement on a 7.1.3 / ABI V9 kernel — sharing in case it's useful:

```
from_all(V9).contains(ResolveUnix) = true;  restrict_self -> FullyEnforced
connect() to an out-of-domain pathname socket     -> EACCES
connect() under a ResolveUnix-granted PathBeneath -> OK
```

Deny and grant paths both work. Bit is `1<<16` per the uapi header, V8 is TSYNC-only (opt-in), V1–V7 unchanged.

l0kod:
Thanks for the PR, but the TSYNC support is almost done (see https://github.com/landlock-lsm/rust-landlock/pull/130) and I'll sent a dedicated one for the UNIX pathname control soon. I'll add a comment here if you're interested to review it.

l0kod:
See #135

PR #132, opened 2026-07-23
Fix flaky CI

l0kod:
See:
* https://github.com/landlock-lsm/rust-landlock/actions/runs/30010904675/attempts/1?pr=132
* https://github.com/landlock-lsm/rust-landlock/actions/runs/30010904675/attempts/2?pr=132

This is related to an UML bug when running on new CPUs. Because the GitHub CI runners use different kind of machines, Linux with UML only failed on the newers. I'll backport the fix in https://github.com/landlock-lsm/landlock-test-tools.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,7 +12,7 @@ env:
   CARGO_TERM_COLOR: always
   RUSTDOCFLAGS: -D warnings
   RUSTFLAGS: -D warnings
-  LANDLOCK_TEST_TOOLS_COMMIT: fad769c39b42183fb2a2e1263fe00dfa5b9f2bda
+  LANDLOCK_TEST_TOOLS_COMMIT: 1dd31a0f7961cd7107b78497859d593849f30008
 
 # Ubuntu versions: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
 
@@ -103,7 +103,7 @@ jobs:
       if: steps.cache_linux.outputs.cache-hit != 'true'
       working-directory: linux
       run: |
-        O=. ../landlock-test-tools/check-linux.sh build_light
+        O=. BUILD_FLAVOR=light ../landlock-test-tools/check-linux.sh build
         mv linux ../linux-${{ fromJSON(matrix.kernel).version}}
 
   ubuntu_24_rust_msrv:


eslerm:
Since #132 is about the UML flakiness and you've already got the virtme-ng path going for the kselftests (`x86-run.sh`), I tried running the *crate* tests under vng/KVM on a plain `ubuntu-24.04` runner to see if it sidesteps the ptrace-fp issue. It does, and it also picks up ABI 9, which the UML matrix can't reach today. Proof-of-concept run on a fork: https://github.com/eslerm/rust-landlock/actions/runs/30066762313

A few things that came out of it:

- KVM is available on the free hosted runners after the usual udev rule (`kvm-ok`: "KVM acceleration can be used"), so no larger or self-hosted runner needed.
- The lib test suite (the same binary `test-rust.sh` runs under UML) passes under vng against three kernels: the runner's own 6.17 (ABI 7), a vng-built mainline `v7.0` (ABI 8), and a vng-built mainline `v7.1` (ABI 9). I read `landlock_create_ruleset(NULL,0,VERSION)` inside each guest and gate on it, so the ABI is really 7/8/9, not `can_emulate` hiding a lower one.
- The `v7.1` rung is ABI 9 (`ResolveUnix`), which the UML matrix can't build today (no `linux-7.1.y` branch, and `kernel_list` bails on it), so it's coverage that isn't reachable right now.

Because the guest kernel runs natively under KVM rather than through UML's ptrace-based FP save/restore, the xstate-size mismatch can't arise, so this also wouldn't need the per-kernel backport. Not suggesting you drop UML, just that the vng path you've already started seems to extend cleanly to the cargo tests and up to v9.

l0kod:
> Since #132 is about the UML flakiness and you've already got the virtme-ng path going for the kselftests (`x86-run.sh`), I tried running the _crate_ tests under vng/KVM on a plain `ubuntu-24.04` runner to see if it sidesteps the ptrace-fp issue. It does, and it also picks up ABI 9, which the UML matrix can't reach today. Proof-of-concept run on a fork: https://github.com/eslerm/rust-landlock/actions/runs/30066762313

Thanks, that's indeed an option but I'd like to stick to UML for quicker tests. I guess at some point a VM might be more convenient but not sure for this specific task yet.

Another reason to use UML is to make the test harness more widely available, without requiring privileged access (e.g. /dev/kvm). I plan to make it more generic for anyone to use it and test Landlock against different kernel versions.

> 
> A few things that came out of it:
> 
> * KVM is available on the free hosted runners after the usual udev rule (`kvm-ok`: "KVM acceleration can be used"), so no larger or self-hosted runner needed.

Good to know. I think that wasn't the case when I started testing this crate a few years ago.

> * The lib test suite (the same binary `test-rust.sh` runs under UML) passes under vng against three kernels: the runner's own 6.17 (ABI 7), a vng-built mainline `v7.0` (ABI 8), and a vng-built mainline `v7.1` (ABI 9). I read `landlock_create_ruleset(NULL,0,VERSION)` inside each guest and gate on it, so the ABI is really 7/8/9, not `can_emulate` hiding a lower one.
> * The `v7.1` rung is ABI 9 (`ResolveUnix`), which the UML matrix can't build today (no `linux-7.1.y` branch, and `kernel_list` bails on it), so it's coverage that isn't reachable right now.

Yes I know, I just wanted to share my work early. I'll push a Linux 7.1 branch when ready.

> 
> Because the guest kernel runs natively under KVM rather than through UML's ptrace-based FP save/restore, the xstate-size mismatch can't arise, so this also wouldn't need the per-kernel backport. Not suggesting you drop UML, just that the vng path you've already started seems to extend cleanly to the cargo tests and up to v9.

I fixed the bug for UML (it's already fixed upstream but not backported), so let's go with that for now.

l0kod:
Update landlock-test-tools to bring improvements and fixes, especially:
- https://github.com/landlock-lsm/landlock-test-tools/commit/f644a251231e: reported result fix;
- https://github.com/landlock-lsm/landlock-test-tools/commit/1dd31a0f7961: UML FP XSTATE fix.

PR #135, opened 2026-07-23
Add ABI v9: ResolveUnix

l0kod:
Add support for Landlock ABI v9.

Depends on https://github.com/landlock-lsm/rust-landlock/pull/136


eslerm:
Thanks for spinning the pathname-socket control out into its own PR. Gave it the same bare-metal pass as #130/#131: built from `pull/135/head` (`6b261dd`) and ran the crate suite on a 7.1.4 / ABI v9 box (x86_64, clean tree). Green, including `consistent_access_fs_rw`.

Since `abi_v9_resolve_unix` grants `ResolveUnix` on `/`, it covers the build/compat path but not the `connect()` deny, so I checked that part directly. On v9, handling `from_all(V9)` and granting everything on the socket's directory except `ResolveUnix`, a `connect()` to a pathname `AF_UNIX` socket gets `EACCES` (the errno for a filesystem access right, where scope uses `EPERM`); add `ResolveUnix` to the grant and it goes through, and not handling it at all leaves it allowed too. So the deny is coming from the right itself, not from just being in a domain, and it was stable across repeated runs. An abstract-namespace `connect()` still goes through under the same wall, so the right stays pathname-scoped and doesn't lean on scope.

I also booted the same host into its 7.0.13 / ABI v8 kernel (ABI from a direct `landlock_create_ruleset(NULL,0,VERSION)` probe, not uname): there the same `deny` ruleset comes back `PartiallyEnforced` and the connect is allowed, i.e. `ResolveUnix` is best-effort dropped and the non-`FullyEnforced` status makes that visible to the caller.

One heads-up on CI: `kernel_list` is red because the workflow now lists `7.1`, but `landlock-lsm/linux` doesn't have a `linux-7.1.y` branch yet, so the `ls-remote` comes back empty and the step hits its `exit 1` (`get_kernels` and the stable job then skip). Probably just sequencing on your end; flagging it since the PR shows red.

l0kod:
Thanks for the tests @eslerm !

